Issue #209, opened 2024-07-09
signal-level caching


gtrevisan:
I believe it would be interesting to first try and benchmark signal-level caching using low-hanging fruit like [Python's own LRU cache decorators](https://docs.python.org/dev/library/functools.html#functools.lru_cache), if applicable.

otherwise, we should take inspiration from Josh's suggestions:
- #173 
- #188

EDIT: currently being tested with mixed results:
- #248

gtrevisan:
> it seems to me that your new tests show an improvement of roughly ~2% on average, and a slight delay at worst. :turtle: 
> I suppose that's not worth the added complication, so we can put this on hold for now. :roll_eyes: 

_Originally posted by @gtrevisan in https://github.com/MIT-PSFC/disruption-py/issues/248#issuecomment-2568036727_

PR #248, opened 2024-08-01
Signal caching

amdecker:
## Problem
The output of the EFIT and physics methods are cached. Individual MDSplus signals are not cached, so we waste time when fetching them.

- https://github.com/MIT-PSFC/disruption-py/issues/209

## Proposed Solution
Modify the `get` methods of `mds.py` to cache their results in a dictionary. 

The key of the dictionary is a tuple of the MDS expression, the tree name, and any arguments. The value is the output of the mds call. 

## Evaluation
(Edit: add reproducible profiling example)
To verify:
Add `profile_script.py` to the disruption-py project directory:
```
#!/usr/bin/env python3

get_shots_data(
    tokamak="cmod",
    shotlist_setting="cmod_ufo", # 122 shots
    retrieval_settings=RetrievalSettings(),
    num_processes=1,
)
```
I used https://github.com/benfred/py-spy for profiling and https://www.speedscope.app/ for viewing the profiles. 

Run it once or twice to prime the mds cache then run on both cache and dev to compare. 
On signal-caching branch: `py-spy record --rate 500 -o signal-caching.speedscope --format speedscope -- python profile_script.py`
On dev branch: `py-spy record --rate 500 -o dev.speedscope --format speedscope -- python profile_script.py`

Results:

With signal caching:
image

Without signal caching:
image

And overall times: signal caching took 55 secs, dev took 67 secs (signal caching is 20% faster).

### DIFF
--- a/disruption_py/core/utils/misc.py
+++ b/disruption_py/core/utils/misc.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from tempfile import mkdtemp
 from typing import List
 
+import MDSplus
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


amdecker:
image
image

gtrevisan:
it seems to me that your new tests show an improvement of roughly ~2% on average, and a slight delay at worst. :turtle: 
I suppose that's not worth the added complication, so we can put this on hold for now. :roll_eyes: 

we'll chat about this next time, then!

PR #271, opened 2024-08-28
Simplify multiprocessing

amdecker:
## Problem
We want prettier logging and progress bars, but there is not a super simple way to integrate `multiprocess.py` with a tqdm-like loader bar. Also, `multiprocess.py` seems unnecessary (@lajz lmk if there's a good reason for this file). 

## Proposed solution
Delete `multiprocessing.py` and replace with Python's builtin multiprocessing `imap`. To integrate a tqdm loader in the future, all we would need to do is replace

```
pool.imap(_execute_retrieval, args)
```
with
```
tqdm(pool.imap(_execute_retrieval, args))
```

## Verification
All tests pass and there is no time difference between the custom logic in `multiprocessing.py` and using `imap`.


gtrevisan:
did you try an extensive benchmark (eg 1,000 shots) to test out edge cases that might be missing from our test set, or simply to put the scheduler under a bit more stress?

disruption_py/workflow.py
```diff
@@ -70,14 +98,11 @@ def get_shots_data(
...
     tokamak = resolve_tokamak_from_environment(tokamak)
+    if database_initializer:
+        database = database_initializer()
+    else:
+        database = get_database(tokamak=tokamak)
 
-    database_initializer = database_initializer or (
-        lambda: get_database(tokamak=tokamak)
-    )
-    database = database_initializer()
```
gtrevisan, line 77: AFAIU there is a difference in how we treat the DB and how we treat the MDSplus server -- can we exploit the same paradigm (object vs initializer) for both?
or is this DB creation necessary because we need to init and resolve the DB to instantiate both `ShotlistSettingParams` and `OutputSettingParams`?

gtrevisan (commented):
I'm a bit confused by the object vs initializer difference for DB and MDSplus, plus how they are then obtained by each process -- but my doubts are probably only marginally related to this PR, and more to the current multiprocessing workflow...

disruption_py/workflow.py
amdecker: We pass the initializer for both to `_execute_retrieval` otherwise there would be pickling issues, but we need the db creation for `ShotlistSettingParams` and `OutputSettingParams`.

gtrevisan (approved):
:crossed_fingers:

PR #310, opened 2024-09-17
Break up configs

amdecker:
## Changes
- Use config.toml files for each tokamak to separate common from individual settings.
- Added a test in `tests/test_config` to ensure the individual settings are accessible


amdecker:
1. Added `~/.config/disruption-py/user.toml`. As with the other Dynaconf toml files, it will override any settings that are strings or numbers, but will merge with existing settings for dicts and lists (https://www.dynaconf.com/merging/)
2. I added a dependency, [toml-sort](https://github.com/pappasam/toml-sort?tab=readme-ov-file#unformatted-unsorted-input). It has the ability to check for errors and fix them like our other linters, so I added it to the make file.
3. LMK what you think about the new structure
4. I generally followed this structure, though I did not add the `machine` level because I think it better follows the structure of the code. Otherwise something like `machine.cmod.inout` suggests `inout` falls under `machine`.

gtrevisan (commented):
I've started the review, but we'll definitely chat about this. :smile: 

- I reverted the typo fix as there is another PR for that,
- I've slightly updated the makefile,
- I have not yet investigate the failure on `v_0`!

PS: for the future, please break up the commits like:
1. modify pyproject.toml,
2. update poetry.lock,
3. apply linting changes.

I feel it'd be cleaner. :wink:

amdecker:
Looks like `v_0` was removed from `test_columns` in https://github.com/MIT-PSFC/disruption-py/pull/320. Just added a commit here to mirror that change. I also double-checked the other column lists in cmod and d3d to make sure they are the same as dev.

gtrevisan (approved):

Issue #292, opened 2024-09-09
Improve logging


amdecker:
## New tools
- [x] [tqdm](https://tqdm.github.io/) because it makes progress bars very easy to implement
    - Both standard logging and loguru can be made compatible with it by adding a custom handler
    - TQDM will be easy to integrate after changes in https://github.com/MIT-PSFC/disruption-py/pull/271
- [x] [loguru](https://loguru.readthedocs.io/en/stable/overview.html)  
    - Benefits over standard logging:
        - Easy formatting of strings, colors
        - Works with multiprocessing with [`enqueue` parameter](https://loguru.readthedocs.io/en/stable/overview.html#asynchronous-thread-safe-multiprocess-safe). The standard logger [does not support](https://docs.python.org/3/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes) multiprocessing
        - We could use their built-in exception [catcher/logger decorator](https://loguru.readthedocs.io/en/stable/overview.html#exceptions-catching-within-threads-or-main)

### Basic Examples log + tqdm
Standard logging
```python
import logging
import time

from tqdm.auto import tqdm


class TqdmLoggingHandler(logging.Handler):
    def emit(self, record):
            msg = self.format(record)
            tqdm.write(msg)
            
            
logger = logging.getLogger("example")
logger.setLevel(logging.INFO)
tqdm_handler = TqdmLoggingHandler()
formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
tqdm_handler.setFormatter(formatter)

logger.addHandler(tqdm_handler)

for i in tqdm(range(10)):
    time.sleep(0.5)
    logger.info(f"Completed iteration {i + 1}")
```

Loguru
```python
import time

from loguru import logger
from tqdm.auto import tqdm

class TqdmLoggingHandler:
    def write(self, message):
        tqdm.write(message, end='')

# Remove the default handler
logger.remove()  

# Add tqdm handler
logger.add(TqdmLoggingHandler(), format="<red>{time}</red> <level>{message}</level>", colorize=True)

x = 0
for i in tqdm(range(10)):
    time.sleep(0.5)
    logger.opt(colors=True).debug("{loop_idx} <green>this is green</> <red>this is red</>", loop_idx=i)
    logger.debug("debug color")
    logger.warning("this is warning color")
```

## Proposed changes to logging content
- [x] Add information about Python version to the beginning of a log file
[...]

PR #319, opened 2024-09-26
Revamp logging with loguru

amdecker:
Implements changes for loguru/logging suggested in 
- https://github.com/MIT-PSFC/disruption-py/issues/292

Levels and their descriptions:
- DEBUG: Used for detailed stack traces. **_Very Verbose_**, the default level when logging to file.
- INFO: Indicates completion of a single physics method.
- SUCCESS: Indicates completion of all physics methods for a single shot. 
- WARNING: Some problem occurred, but it is probably not too serious. 
- ERROR: A serious error occurred, e.g. SQL query failing.

New features
- Now we have colors in the console logs
- Time elapsed is rounded to milliseconds
- A success/warning log for each shot
- A success/warning log at the end
- tqdm progress bar
- dynamic adjustment of log level based on number of shots

image


disruption_py/settings/log_settings.py
```diff
@@ -114,4 +117,15 @@ def setup_logging(self):
...
+        package = "disruption_py"
+        commit = get_commit_hash()
+        logger.log(
+            "SUMMARY",
+            f"> {package} "
+            + f"v{importlib.metadata.version(package)} "
+            + f"# {commit} "
+            + f"/ {os.getenv('USER')}@{os.uname().nodename}",
+        )
+        logger.debug(f"https://github.com/MIT-PSFC/disruption-py/commit/{commit}")
```
gtrevisan, line 130: @amdecker we could populate this URL using the repository field from pyproject, but I could not find an easy way...

amdecker: Maybe you already saw this, but https://docs.python.org/3/library/tomllib.html was added in 3.11. When we drop 3.10, it could be worth looking into because then we could to something like 
```
with open("pyproject.toml", "rb") as f:
    data = tomllib.load(f)
repo = data["tool"]["poetry"]["repository"]
```

disruption_py/core/physics_method/runner.py
```diff
@@ -237,19 +232,17 @@ def populate_method(
...
         result = {col: [np.nan] for col in bound_method_metadata.columns}
 
     physics_method_params.logger.info(
-        "[Shot %s]: Completed %s, time_elapsed: %s",
-        physics_method_params.shot_id,
-        name,
-        time.time() - start_time,
+        "{t:.3f}s : {name}",
+        name=name,
+        t=time.time() - start_time,
```
gtrevisan, line 245: how come `start_time` is passed as an arg, rather than getting created as a first thing by the method itself?

disruption_py/workflow.py
```diff
@@ -121,26 +125,39 @@ def get_shots_data(
...
                     OutputSettingParams(
                         shot_id=shot_id,
                         result=shot_data,
                         database=database,
                         tokamak=tokamak,
-                        logger=logger,
                     )
                 )
+        total = len(shotlist_list)
+        percent_success = round(num_success / total * 100, 2)
```
gtrevisan, line 150: was there a reason for rounding? I removed the rounding in the other instance when I modified it.

disruption_py/core/physics_method/runner.py
amdecker: Here, the start time means the start time of running the physics method at the level of the shot. We log the elapsed time since the start of getting data for the shot at the completion of each physics method.

Here's an alternate way to structure these time logs:
- Each shot logs the time to grab all data for the shot in our that shot's success/warning/error log eg `#1150805014 | Success! Most parameters have data: 62/63 (98.41%) in 1s`
- Each physics method logs only the time it took to complete that one physics method

Let me know how that sounds

gtrevisan: okay, we'll go through this tomorrow! thanks.

disruption_py/core/utils/math.py
```diff
@@ -175,8 +187,9 @@ def gaussian_fit(*args):
     coeffs : array
         The coefficients of the fit.
     """
-
-    coeffs, *_ = curve_fit(gauss, *args)
+    with warnings.catch_warnings():
```
gtrevisan, line 190: I did not do this myself because I thought that it would have been safer to filter class (OptimizeWarning) but _also_ message.
at present scipy.curve_fit does raise only covariance-based OptimizeWarnings, but what if they add anything new in that class? we wouldn't ever see those warnings again.
hmm... is there a quick and easy way to filter message, too?

disruption_py/core/retrieval_manager.py
```diff
@@ -143,10 +136,7 @@ def shot_setup(
             )
             return physics_method_params
         except Exception as e:
-            self.logger.info(
-                "[Shot %s]: Caught failed to setup shot, cleaning up tree manager.",
-                shot_id,
-            )
+            logger.error(shot_log_msg(shot_id, f"Failed to setup shot! {e}"))
```
gtrevisan, line 139: I added the exception in-line as I believe the actual stack trace does not get added to the log file -- would it make sense to do something different here?

disruption_py/inout/mds.py
```diff
@@ -52,13 +50,43 @@ def get_shot_connection(self, shot_id: int):
...
+
+    def wrapper(*args, **kwargs):
+        self = args[0]
+        if len(args) > 1:
+            path = args[1]
+        else:
+            path = kwargs.get("path", None)
+        try:
+            return func(*args, **kwargs)
+        except MDSplus.mdsExceptions.TreeFOPENR:
```
gtrevisan, line 67: I added `TreeFOPENR` and switched order of the subsequent two exceptions to have them ordered hierarchically.

disruption_py/inout/mds.py
```diff
@@ -52,13 +50,43 @@ def get_shot_connection(self, shot_id: int):
...
+            nick = kwargs["tree_name"]
```
gtrevisan, line 68: is there a better way to accomplish this? feel free to improve.
hopefully `tree_name` always has value...

disruption_py/inout/mds.py
```diff
@@ -52,13 +50,43 @@ def get_shot_connection(self, shot_id: int):
...
+            tree = self.get_tree_name_of_nickname(nick)
+            raise MDSplus.mdsExceptions.TreeFOPENR(
+                f"Tree not found. Nick: {nick}, Tree: {tree}"
```
gtrevisan, line 71: I thought for this specific exception that the nickname of the tree would have been useful, LMK otherwise.

disruption_py/workflow.py
```diff
@@ -95,7 +100,8 @@ def get_shots_data(
     Any
         The value of OutputSetting.get_results. See OutputSetting for more details.
     """
-    (log_settings or LogSettings()).setup_logging()
+    log_settings = log_settings or LogSettings()
```
gtrevisan, line 103: could you please add a case where the user just specifies a string (eg "INFO" or "WARNING") to `get_shots_data`, and it's the framework that instantiates a LogSettings object with that default console log level?

disruption_py/core/utils/misc.py
```diff
@@ -132,3 +138,79 @@ def get_commit_hash() -> str:
...
+
+    # create temporary top folder
+    top = os.path.join("/tmp", os.getenv("USER"), "disruption-py")
+    Path(top).mkdir(parents=True, exist_ok=True)
+
+    # create temporary sub folder
+    return mkdtemp(prefix=time.strftime("%Y%m%d-%H%M%S-"), dir=top)
+
+
+def shot_log_msg(shot_id: int, message: str):
```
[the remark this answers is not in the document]
amdecker: I looked into it for a bit. `shot_log_msg` could return a dictionary 
```
{"_Logger__message": "#{shot_id} | {message}",
        "shot_id": shot_id, 
        "message": message}
``` 
but we would need to always unpack it: `logger.info(**shot_log_msg(shot_id, "this is a message")` which is annoying. 

I'll leave it as is because the lazy loading won't give a huge performance boost.

gtrevisan (approved):
fantastic, this looks great! many thanks :pray:

PR #332, opened 2024-10-18
Fix CMOD EFIT nodenames

yumouwei:
# Implemented changes

- Corrected nodenames in `efit_cols` and `cmod/physics.py`:
  - li -> ali
  - q0 -> qqmagx
  - qstar -> qsta
  - q95 -> qpsib
  - aminor -> aout/100 (aout: [cm])
  - rmagx -> rmagx/100 (rmagx: [cm])
  - chisq -> tsaisq
  - area -> areao/1e4 (areao: [cm^2])
  - zmagx -> zmagx/100 (zmagx: [cm])

- Removed now-unnecessary unit conversions in `cmod/physics.py`
- Removed `efit_cols_pre_2000` and its references in `cmod/efit.py`, as all the nodenames in `efit_cols` have been corrected & match what were in `efit_cols_pre_2000`.

# Remaining issues to be addressed

## 1. beta_n

- The existing nodename `\efit_aeqdsk:betan` is an expression `Build_Signal(\EFIT_AEQDSK:BETAT / \EFIT_AEQDSK:IN, *, \EFIT_AEQDSK:TIME)`. `IN` is the normalized current defined as `Ip[MA]/(a*Bt)`. This expression of beta_n isn't exactly correct as it should be calculated using the total beta instead of beta_t. Also this node is only available for post-2000 shots.

image

- In addition, there's another block of codes under `get_efit_parameters` that calculates beta_n for pre-2000 shots. However that code is also wrong as it calculates the total beta and forget to normalize it by IN.

https://github.com/MIT-PSFC/disruption-py/blob/8888f3f601b03cf922d8053beef93e10801d2fe8/disruption_py/machine/cmod/efit.py#L133-L139

## 2. v_surf

- disruption-py currently calculates `v_surf = gradient(SSIBRY)*2pi` where `SSIBRY` is the poloidal flux at the plasma boundary. We need to make sure this is the correct expression.

https://github.com/MIT-PSFC/disruption-py/blob/8888f3f601b03cf922d8053beef93e10801d2fe8/disruption_py/machine/cmod/efit.py#L109-L117

- There is also a node `\efit_aeqdsk:vsurfa` which is available for all shots but equals 0 for every shot.
- Then there's `efit_aeqdsk:vsurf` which is an alias for `vsurfa` for post-2000 shots.

## 3. v_loop

- For pre-2000 shots, disruption-py calculates `v_loop = gradient(SSIMAG)*2pi` where SSIMAG is the poloidal flux at the magnetic axis. I'm not sure why this is needed as `vloopt` is available for both pre-and-post-2000 shots. The results also don't match `vloopt`.

https://github.com/MIT-PSFC/disruption-py/blob/8888f3f601b03cf922d8053beef93e10801d2fe8/disruption_py/machine/cmod/efit.py#L122-L131

image

# References

- https://github.com/MIT-PSFC/disruption-py/pull/47
- https://github.com/MIT-PSFC/disruption-py/pull/185
- [CMOD wiki: EFIT variables](https://cmodwiki.psfc.mit.edu/index.php/EFIT_variables)
- [NSTX: EFIT Signals](https://nstx.pppl.gov/nstx/Software/Applications/EFIT_Signals.htm)


disruption_py/machine/cmod/efit.py
```diff
@@ -133,12 +116,9 @@ def get_efit_parameters(params: PhysicsMethodParams):
             print("unable to get V_surf")
             efit_data["v_surf"] = np.full(len(efit_time), np.nan)
 
-        # For shots before 2000, adjust units of aminor, compute beta_n and v_loop
+        # For shots before 2000, compute beta_n and v_loop
         if params.shot_id <= 1000000000:
 
-            # Adjust aminor units
-            efit_data["aminor"] = efit_data["aminor"] / 100  # [cm] to [m]
```
gtrevisan, line 140: this was probably a bug, but let's double check because `aminor` was already in meters!
if the code downstream from this line was expecting it in meters, it was effectively getting it in... hundreds of meters. :confused:

disruption_py/machine/cmod/physics.py
```diff
@@ -685,10 +685,10 @@ def get_kappa_area(params: PhysicsMethodParams):
             A dictionary containing the calculated "kappa_area".
         """
         aminor = params.mds_conn.get_data(
-            r"\efit_aeqdsk:aminor", tree_name="_efit_tree", astype="float64"
+            r"\efit_aeqdsk:aout/100", tree_name="_efit_tree", astype="float64"
```
gtrevisan, line 688: at least in the scope of this method, let's keep adding comments for units...
I _might_ prefer having them before the line, for brevity, eg:
```
# [m] -- or even, more explicitly -- r0 [m]
r0 = params.mds_conn.get_data(r"\efit_aeqdsk:rmagx/100", tree_name="_efit_tree")
```
vs
```
r0 = params.mds_conn.get_data(
            r"\efit_aeqdsk:rmagx/100", tree_name="_efit_tree"
)  # [m]
```
your choice though!

gtrevisan (commented):
@yumouwei the renames LGTM, but I still have to take a look at the bigger points you raised.

@crea-psfc do we want to add `areao` to the quantities fetched by this EFIT method?

@amdecker is there any implication between these two paradigms for your upcoming  signal caching implementation?
```
a = mds.get_data(r"\efit_aeqdsk:aout/100")
# vs.
a = mds.get_data(r"\efit_aeqdsk:aout") / 100
```

disruption_py/machine/cmod/physics.py
amdecker: re
```
a = mds.get_data(r"\efit_aeqdsk:aout/100")
# vs.
a = mds.get_data(r"\efit_aeqdsk:aout") / 100
```
We would only be caching the expression, so in the first case we would cache the entire expression `"\efit_aeqdsk:aout/100"`. If we always divide `aout` by 100 any time we use it, it would be better to keep it in the expression to save us on that division. If we are using `aout` and not always diving by 100 `mds.get_data(r"\efit_aeqdsk:aout") / 100`. Though I don't know how much time caching a single division operation will really help us.

gtrevisan: it should be reasonable that reading `r"\efit_aeqdsk:aout/100"` or `r"\efit_aeqdsk:aout"` will be _two separate_ cache lookups, but I was wondering whether you were doing anything more specific than using the node path as the cache key -- maybe you had already planned for trivial parsing of major operations? eg `*`, `/`, `+`, `-`.

yumouwei: I added units to most of the individual MDS calls in `cmod/efit.py` and `cmod/physics.py`. 
@gtrevisan How should we add units to the lines in `efit_cols`?

disruption_py/machine/cmod/efit.py
yumouwei: So before my changes:

- For pre-2000 shots, the function gets a_minor from '\efit_aeqdsk:aout' in [cm) and then divides the signal by 100 ([cm] -> [m])
- For post-2000 shots, the function gets a_minor from '\efit_aeqdsk:aminor' which is already in [m]

It's somewhat unnecessary but it worked.

disruption_py/machine/cmod/physics.py
gtrevisan: naa, this is fine for now -- in the future we might want to use something like pint, or maybe try and fetch the units from MDSplus itself.

crea-psfc:
Thank you @yumouwei for the thorough review. Generally, all pre-2000 code blocks need to be thoroughly reviewed and validated. Regarding the specific points raised:

1. Let's consult with Bob Granetz to determine the correctness of `beta_n`. 
2. I believe the one you linked is the correct expression for `vsurf` as C-Mod had one big poloidal Rogowski loop. Bob Granetz is the one that can solve this point as well.
3. same as above, let's raise these points to Bob's attention to determine best course of action.

@gtrevisan there are certainly edge cases for DIII-D as well, but we never went into similar detailed analysis of historical data. I am aware of a strong bifurcation in the availability of the n=1 bradial signal. I talked to @yumouwei at the time, he should be aware. Regarding `areao`, sure, let's add it to the list of retrieved signals from EFIT. Right now it can be reconstructed via differences between `kappa` and `kappa_area`.

yumouwei:
Following the discussion on 11/4/24, we decided to investigate & address the beat_n, v_surf, & v_loop computations in a separate PR. For now this PR is ready for review.

gtrevisan (approved):

PR #333, opened 2024-10-22
Add script for dependency updates

gtrevisan:
add a script to semi-automatically update dependencies:
- major updates flagged for user intervention,
- minor updates automatically merged onto pyproject.toml,
- patch updates automatically merged onto poetry.lock.

e.g.
```
bash scripts/poetry-update.sh -n
```

if anybody has thoughts on this, or suggestions, feel free to reach out!
IMHO this strikes a good balance between staying somewhat-up-to-date with our deps, and keeping a widened but reasonable dep list.

### DIFF
--- a/scripts/poetry-update.sh
+++ b/scripts/poetry-update.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+set -e
+
+PYPROJECT="$(dirname "$0")/../pyproject.toml"
+TMPF=$(mktemp)
+
+R="\e[31m"
+Y="\e[33m"
+G="\e[32m"
+B="\e[34m"
+U="\e[4m"
+X="\e[0m"
+
+{
+   echo main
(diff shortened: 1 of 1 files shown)


amdecker (approved):
colors! 🌈

Logic makes sense for keeping things up-to-date. Let's hope no one puts a bug in their minor update!

gtrevisan:
I usually [create a PR to update the deps](https://github.com/MIT-PSFC/disruption-py/pulls?q=is%3Apr+label%3Adependencies), so it should be apparent if stuff does not work -- in any case, the minor version locking is just a convention, the actual installed version is governed by the lockfile, and that does track patch versions, too.

I don't *push* all the patch versions every time, but your poetry is independent of that, eg you can run `poetry update` whenever you want.
(worst case is you'll have some git diffs in your local lockfile.)

PR #343, opened 2024-11-06
Compute beta normalized for both old and new C-MOD shots

yumouwei:
# Original problems

- `\efit_aeqdsk:betan` is undefined for pre-2000 shots.
- There's a separate `betan` computation in `cmod/efit.py` for pre-2000 shots written by someone, but the calculation is wrong.

# Implemented changes

- Adding `get_beta_normalized` function in `cmod/physics.py`. This method takes in beta_t, ip, aminor, & btor data from EFIT and computes beta_n using the same formulation as `\efit_aeqdsk:betan`. The output agrees with `\efit_aeqdsk:betan` for post-2000 shots.
- Remove existing `betan` fetching & computation in `cmod/efit.py`.
- Update CMOD output shape in `examples/efit.py`.

### DIFF
--- a/disruption_py/machine/cmod/efit.py
+++ b/disruption_py/machine/cmod/efit.py
@@ -114,7 +114,7 @@ class CmodEfitMethods:
             print("unable to get V_surf")
             efit_data["v_surf"] = np.full(len(efit_time), np.nan)
 
-        # For shots before 2000, compute beta_n and v_loop
+        # For shots before 2000, compute v_loop
         if params.shot_id <= 1000000000:
 
             # Get data for v_loop --> deriv(\ANALYSIS::EFIT_SSIMAG)*$2pi (not totally
--- a/disruption_py/machine/cmod/efit.py
+++ b/disruption_py/machine/cmod/efit.py
@@ -30,7 +30,6 @@ class CmodEfitMethods:
     """
 
     efit_cols = {
-        "beta_n": r"\efit_aeqdsk:betan",
         "beta_p": r"\efit_aeqdsk:betap",
         "kappa": r"\efit_aeqdsk:eout",
         "li": r"\efit_aeqdsk:ali",
@@ -59,7 +58,6 @@ class CmodEfitMethods:
             *efit_derivs.keys(),
             "v_surf",
             "v_loop_efit",
-            "beta_n",
         ],
         tokamak=Tokamak.CMOD,
     )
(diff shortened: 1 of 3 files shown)


disruption_py/machine/cmod/efit.py
```diff
@@ -130,14 +128,6 @@ def get_efit_parameters(params: PhysicsMethodParams):
                 print("unable to get v_loop_efit")
                 efit_data["v_loop_efit"] = np.full(len(efit_time), np.nan)
 
-            # Compute beta_n
```
gtrevisan, line 133: there was another comment about `beta_n` just before the `if` statement acting on pre-2000 shots.

disruption_py/machine/cmod/physics.py
```diff
@@ -1754,6 +1754,64 @@ def get_sxr_data(params: PhysicsMethodParams):
         sxr = interp1(t_sxr, sxr, params.times)
         return {"sxr": sxr}
 
+    @staticmethod
+    @physics_method(
```
gtrevisan, line 1758: you should be able to have these decorator arguments in a single line without black complaining.

disruption_py/machine/cmod/physics.py
```diff
@@ -1754,6 +1754,64 @@ def get_sxr_data(params: PhysicsMethodParams):
...
+        for the plasma current and toroidal field data.
+
+        Parameters
+        ----------
+        params : PhysicsMethodParams
+            The parameters containing the MDSplus connection, shot id and more.
+
+        Returns
+        -------
+        beta_n: the normalized beta given in percentage.
```
gtrevisan, line 1783: I'm not sure how we prepared the other docstrings, but we return a dict which holds betan, not betan itself.

gtrevisan (changes requested):
LGTM! minor comments.

gtrevisan:
@crea-psfc is it worth analyzing the same for DIII-D, as well?

crea-psfc:
not aware of similar issues, but it's worth it to do a quick check. IMHO, EFIT is better maintained at DIII-D

disruption_py/machine/cmod/physics.py
yumouwei: I tried but it doesn't seem to work.

gtrevisan: I tried it and it did work. :smile:

gtrevisan (approved):

PR #344, opened 2024-11-07
Wei/cmod vsurf vloop

yumouwei:
# Current problems

## v_surf
v_surf computation in `cmod/efit.py` has the wrong sign.

## v_loop
There are two nodes that stores the loop voltage data: 

1. `\analysis:top.mflux:v0`: this node has a higher sampling rate, is being used in `p_oh` calculation, is NOT available for pre-2000 shots.
2. `\efit_aeqdsk:vloopt`: this node samples at the EFIT time base, appears to match `mflux:v0` for post-2000 shots, is available for pre-2000 shots but has higher noise level.

In addition, there's a separate computation in `cmod/efit.py` for pre-2000 shots, using `v_loop_efit = deriv(SSIMAG)*2pi`. There are two problems:
1. It's missing a minus sign
2. The results don't appear to match `vloopt`.

image
(shot 1150805012)

@andrew-maris can you comment on this? I'm not sure if this was add by you or someone else.

# Implemented changes

- Corrected v_surf computation and moved it to `cmod/physics.py` as `get_surface_voltage` function.


gtrevisan:
@yumouwei this PR looks ready for review to me. are you working on anything else here?

andrew-maris (commented):
Good catch on the sign of vsurf, I talked to Bob and this is indeed correct. He said the signal might not be trustworthy, but that's how he would compute it.

I see this issue also touches on v_loop, are you planning to make any changes regarding that?

Bob and I dug into this for one shot (the last C-Mod shot 116930043). mflux:v0 is an estimate of vloop based on weighted measurements of the voltage outside the plasma by Ian. This appears noisier, but that's just due to the far higher sampling rate. We found vloopt seemed to be more trustworthy than deriv(ssimag)*-2pi because it quantitatively matches mflux:v0. It also follows a trend we would expect: noticeably greater magnitude during ramp-up.

Obviously this was just a look at one shot, but if you plan to address v_loop, I hope this helps.

yumouwei:
Thanks for the comment. In that case I'd 
1. Keep `v_loop_efit` (`efit_aeqdsk:vloopt`) and `v_loop` (`mflux:v0`)
2. Remove the `(-1)*deriv(SSIMAG)*2pi` function from `cmod/efit.py`

andrew-maris (approved):
Looks good to me, nice work 👍

gtrevisan (approved):

PR #356, opened 2024-12-03
Updated get_surface_voltage computation

yumouwei:
Changed `v_surf` computation from:
```
v_surf = -2pi*deriv(\efit_geqdsk:ssibry)
```
to:
```
v_surf = -2pi*deriv(\efit_aeqdsk:sibdry)
```
per suggestion from Stephen Wolfe.
Now `v_surf` will always have the same sign as `ip` regardless of the sign of `ip`

image

(A positive ip case)

image

(A negative ip case)


gtrevisan (approved):
thank you @wolfe-psfc for the suggestion, and @yumouwei for the implementation!

PR #358, opened 2024-12-04
Fix commit hash test

amdecker:
## Problem

Currently, on dev, running pytest fails:
```
tests/test_retrieval_settings.py::test_cache_setting_prev_output[.csv] - AssertionError: Attributes of DataFrame.iloc[:, 62] (column name="commit_hash") are different
```

But running just that one test file succeeds:
```
pytest tests/test_retrieval_settings.py
```

## Proposed Solution
Modify the `change_directory` decorator to always change back to the current directory after changing to the `/tmp` directory.


gtrevisan (approved):
do we need to change it back at all?

PR #359, opened 2024-12-10
Fix nan-multiplication error in get_peaking_factors for C-MOD

yumouwei:
Added a `np.errstate` context manager to avoid a `RuntimeWarning: invalid value encountered in multiply` error in `get_peaking_factors` computation.

Previously failing shot: 1150728005

### DIFF
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -1173,13 +1173,13 @@ class CmodPhysicsMethods:
         # Fetch data
         # Get EFIT geometry data
         z0 = params.mds_conn.get_data(
-            r"\efit_aeqdsk:zmagx/100", tree_name="_efit_tree", astype=np.float64
+            r"\efit_aeqdsk:zmagx/100", tree_name="_efit_tree", astype="float64"
         )  # [m]
         kappa = params.mds_conn.get_data(
-            r"\efit_aeqdsk:kappa", tree_name="_efit_tree", astype=np.float64
+            r"\efit_aeqdsk:kappa", tree_name="_efit_tree", astype="float64"
         )  # [dimensionless]
         aminor, efit_time = params.mds_conn.get_data_with_dims(
-            r"\efit_aeqdsk:aout/100", tree_name="_efit_tree", astype=np.float64
+            r"\efit_aeqdsk:aout/100", tree_name="_efit_tree", astype="float64"
         )  # [m], [s]
         bminor = aminor * kappa
 
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -1181,7 +1181,9 @@ class CmodPhysicsMethods:
         aminor, efit_time = params.mds_conn.get_data_with_dims(
             r"\efit_aeqdsk:aout/100", tree_name="_efit_tree"
         )  # [m], [s]
-        bminor = aminor * kappa
+        # Avoid edge case error caused by NaN * 0
+        with np.errstate(invalid="ignore"):
+            bminor = aminor * kappa
 
         # Get Te data and TS time basis
         node_ext = ".yag_new.results.profiles"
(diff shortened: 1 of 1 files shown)


gtrevisan:
can you provide a snippet to reproduce?

yumouwei:
Minimum example

```
shotno = 1150728005

shot_settings = RetrievalSettings(
    run_columns=['ip', 'ne_peaking', 'te_peaking', 'pressure_peaking'],
    only_requested_columns=True,
)
result = get_shots_data(
    shotlist_setting=[shotno],
    retrieval_settings=shot_settings,
    output_setting='dataframe',
)
```

gtrevisan:
got it now, thanks.

I feel we could probably choose a more general path, here.
what about something more explicit like:
```python
bminor = np.where(np.isnan(aminor) | np.isnan(kappa), np.nan, aminor * kappa)
```
? (or, `np.isnan --> np.isfinite`, etc)

@amdecker didn't we encounter other similar cases in revamping the methods? what did we do there?

amdecker:
This was interesting! There are two types of nans: signalling nans and silent nans. When we do an operating with signalling nans (eg nan * 1) it will give us a warning. If we do an operation with silent nans, the operation will give nan.  It seems like MDSPlus uses signalling nans which is why there is this error. To get rid of the error, we could turn them into silent nans (the numpy default) by converting the arrays to np.float64 using the `astype` parameter in the mds call.

```python
kappa = params.mds_conn.get_data(
            r"\efit_aeqdsk:kappa", tree_name="_efit_tree", astype=np.float64
        )  # [dimensionless]
        aminor, efit_time = params.mds_conn.get_data_with_dims(
            r"\efit_aeqdsk:aout/100", tree_name="_efit_tree", astype=np.float64
        )  # [m], [s]
```

gtrevisan:
yes, our utility functions do support an `astype` argument (which BTW can be a simple string, eg `"float64"`), that's why we did not see (an overabundance of) similar errors for C-MOD or DIII-D -- cool!

gtrevisan (approved):
@yumouwei @AlexSaperstein please verify this does indeed fix your issue, then LMK.

@amdecker once again, great investigation! :mag:

yumouwei:
@AlexSaperstein From your screenshots it looks like both of these bugs occurred in `get_peaking_factor` calculation (lines 1184 & 1085). I'll take a look at the second bug later today.

gtrevisan:
to be clear, a "crash" means that the workflow yields an error and stops -- that would be a nasty bug.

if, instead, some of your parameter methods fail but the workflow collects the respective `np.nan` values and continues, then everything is normal -- you'd still an exception in the logs though.

which is it? also, please provide a minimal reproducible example, as usual. thanks!

AlexSaperstein:
It's definitely "crashing" by your definition. it ends the whole process when it hits this error.

It looks like the locations of the log files is given at the start of the data-fetching process. The last crash was logged at

`/tmp/saperstein/disruption-py/20241212-115259-ty339mle/output.log`

gtrevisan:
oh wait, I think I now remember why this happens -- we coded it in such a way that bugs will _actually be blocking_, so that users would flag them, as opposed to overbroad exception catching which would mean no bugs would be flagged and thus closed. :bug: 

also, MRE:
```python
shot_data = get_shots_data(
    shotlist_setting=[1051123016],
    retrieval_settings=RetrievalSettings(
        run_tags=[], run_methods=["get_peaking_factors"]
    ),
    log_settings="DEBUG",
)
```

yumouwei:
Regarding the second bug, it looks like it's a overflow error caused by multiplying two float32 numbers. Casting both of them into float64 seems to fix the bug.

image

I added `astype='float64'` to the MDS calls for fetching `ts_te` and `ts_ne` and it seems to solve the problem.

AlexSaperstein:
Great, the most recent fix seems to have worked! All of the shots I've had to black-list so far (with the exception of 1120215001, which doesn't have an efit18 tree for some reason) are now working!

gtrevisan:
@AlexSaperstein please test and report -- it's more general and hopefully robust:
- #364

PR #363, opened 2024-12-12
Tweak logging levels and colors

gtrevisan:
- add a new `VERBOSE` level above `DEBUG` but below `INFO` (eg: for output of physics methods),
- tweak some colors, 
- downgrade some logging lines.

to do:
- [x] is there a way to call `logger.verbose("stuff")` as opposed to `logger.log("VERBOSE", "stuff")`?
- [x] it looks like we are setting up the logger multiple times, and that errors out.

### DIFF
--- a/disruption_py/settings/log_settings.py
+++ b/disruption_py/settings/log_settings.py
@@ -36,8 +36,9 @@ class LogSettings:
         By default, a log file will be created in a temporary folder.
     file_log_level : str
         Logging level for the log file (default is "DEBUG").
-        Possible values are: "TRACE", "DEBUG", "INFO", "SUCCESS", "WARNING", "ERROR",
-        "CRITICAL". See https://loguru.readthedocs.io/en/stable/api/logger.html#levels
+        Possible values are:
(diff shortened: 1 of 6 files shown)


disruption_py/settings/log_settings.py
```diff
@@ -150,12 +144,30 @@ def setup_logging(self):
...
+            logger.level("VERBOSE", color="<dim>")
+        except ValueError:
+            logger.level("VERBOSE", color="<dim>", no=15)
+        logger.level("INFO", color="")
+        logger.level("SUCCESS", color="<green>")
+        logger.level("WARNING", color="<yellow>")
+        logger.level("ERROR", color="<red>")
+        # Bind the verbose level to the class so it can be used in any file even
+        # after changing the logger instance
+        logger.__class__.verbose = partialmethod(logger.__class__.log, "VERBOSE")
```
gtrevisan, line 163: ah, so THAT's what I was missing! :mag:

amdecker:
I double checked and looks like all the info logs in the physics methods are now verbose!

gtrevisan:
a lot of them are probably either WARNING or DEBUG, rather than INFO, but we can fix it in the future as we try this new level out. merging! thanks!

Issue #184, opened 2024-06-26
finer control on numpy warnings for C-mod


gtrevisan:
we should remove the overbroad filter warning for all C-mod parameter methods at the module level:
https://github.com/MIT-PSFC/disruption-py/blob/43a6df3307a13b2707de979ec29e75e5388ba0a4/disruption_py/shots/parameter_methods/cmod/basic_parameter_methods.py#L27
and implement finer checks in each of the methods for any numpy warning, eg #170.

gtrevisan:
closed with:
- #373

PR #364, opened 2024-12-12
Cast data to float64 by default

gtrevisan:
in order to avoid these pesky nan-related issues:
- cast all _data_ calls to float64 by default,
- remove explicit assignments from physics methods,
- introduce overrides for strings/names.

[rationale by Amos](https://github.com/MIT-PSFC/disruption-py/issues/359#issuecomment-2537071992):
> There are two types of nans: signaling nans and silent nans. When we do an operation with signaling nans (e.g. nan * 1) it will give us a warning. If we do an operation with silent nans, the operation will give nan.  It seems like MDSPlus uses signaling nans which is why there is this warning. To get rid of the warning, we could turn them into silent nans (the numpy default) by converting the arrays to np.float64 using the `astype` parameter in the mds call.

supersedes:
- #359

### DIFF
--- a/disruption_py/inout/mds.py
+++ b/disruption_py/inout/mds.py
@@ -165,7 +165,7 @@ class MDSConnection:
         self,
         path: str,
         tree_name: str = None,
-        astype: str = None,
+        astype: str = "float64",
         arguments: Any = None,
     ) -> np.ndarray:
         """
@@ -177,7 +177,7 @@ class MDSConnection:
             MDSplus path to record.
         tree_name : str, optional
             The name of the tree that must be open for retrieval.
-        astype : str, optional
+        astype : str, optional, default = "float64"
             The data type for explicit casting.
         arguments : Any, optional
             Arguments for MDSplus TDI Expression. Default None.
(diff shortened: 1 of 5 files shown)


yumouwei:
I checked the 3 shot numbers mentioned in #359 and they all worked. 

@AlexSaperstein can you run your get_data script with the previous seed number and see if you run into any bugs again?

AlexSaperstein:
> I checked the 3 shot numbers mentioned in #359 and they all worked.
> 
> @AlexSaperstein can you run your get_data script with the previous seed number and see if you run into any bugs again?

yup! seems to have worked just fine without having to blacklist any shots!

yumouwei (approved):

gtrevisan:
thanks.

I'll wait for Amos' thoughts on this and on:
- #184

can we remove that line, as well?

amdecker (approved):
Cool! As long as users know the values are not precise for all decimal places

amdecker:
> thanks.
> 
> I'll wait for Amos' thoughts on this and on:
> 
>     * [finer control on numpy warnings for C-mod #184](https://github.com/MIT-PSFC/disruption-py/issues/184)
> 
> 
> can we remove that line, as well?

re the warnings, I removed the line and got shot data for 3K shots on cmod. No new warnings appeared.

gtrevisan:
ok, I will push a separate PR for that.

PR #367, opened 2024-12-23
Improve get_power and get_ohmic_parameters for C-MOD

yumouwei:
# Original problems

1. `p_lh` and `p_rad` data stored in the MDSplus trees have shorter time bases than EFIT for a few shots (e.g. 1120208014 and 1050725002). This causes disruption-py to return nans for these signals (plus `p_input` and `radiated_fraction` which use these signals in their computation) in the missing time period, subsequently causing the ONW module to give bad predictions during these periods.
2. `get_ohmic_parameters` method fails for a few shots (e.g. 1050608005), which also causes `get_power` to fail.

image
image

# Implemented changes

1. Added `interp1(..., fill_value=0)` for `p_lh`, `p_icrh`, and `p_rad` in `_get_power`. It should be safe to assume that these auxiliary heating powers were turned off during the periods where there were no data stored in the tree.
2. Added a try-except block in `get_ohmic_parameters` to return `nans` for `v_loop` and `p_ohm`. 
3. *Modified `_get_power` so that the `p_input` and `radiated_fraction` computations will take `p_ohm=0` even if `get_ohmic_parameters` fails.
    - `p_ohm` and `v_loop` should not be 0 even if there's no data stored in that node in a shot. I want to see there's a better way to do this (see TODO below)
    - This does not affect disruption-py's outputs for these two signals. They will still be arrays of nans when `get_ohmic_parameters` fails.

(Before -- dev)
image

(After)
image

# TODO

- [x] Should we switch to using `\efit_aeqdsk:vloopt` (i.e. `v_loop_efit`), or to use it as a backup in case there's no data stored in `mflux:v0`? @crea-psfc 
- [x] Check error catching.
- [x] Add `tau_rad` from ONW_DEV modules to the `get_power` method.

# Related PR(s)

- #344

### DIFF
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -536,10 +536,12 @@ class CmodPhysicsMethods:
         return output
 
     @staticmethod
-    def _get_power(times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm):
+    def _get_power(
+        times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm, wmhd, efit_time
+    ):
         """
-        Calculate the input and radiated powers, and then calculate the
-        radiated fraction.
+        Calculate the input and radiated powers, then calculate the
+        radiated fraction and radiation confinement time.
 
         Parameters
         ----------
@@ -559,12 +561,18 @@ class CmodPhysicsMethods:
             The time array corresponding to radiated power.
         p_ohm : np.ndarray
             The ohmic heating power.
+        wmhd : np.ndarray
+            The total plasma energy.
+        efit_time : np.ndarray
+            The EFIT time base
 
         Returns
         -------
         dict
             A dictionary containing the calculated power values, including
-            "p_rad", "dprad_dt", "p_lh", "p_icrf", "p_input", and "radiated_fraction".
+            "p_rad", "dprad_dt", "p_lh", "p_icrf", "p_input", "radiated_fraction", and "tau_rad".
+
+        Last major update by William Wei on 1/6/2025
         """
         if p_lh is not None and isinstance(t_lh, np.ndarray) and len(t_lh) > 1:
             p_lh = interp1(t_lh, p_lh * 1.0e3, times, fill_value=0)
@@ -597,9 +605,14 @@ class CmodPhysicsMethods:
         if all(np.isnan(p_ohm)):
             p_ohm = np.zeros(len(times))
 
+        # Calculate radiated fraction
         p_input = p_ohm + p_lh + p_icrf
         rad_fraction = p_rad / p_input
         rad_fraction[rad_fraction == np.inf] = np.nan
+
+        # Calculate radiation confinement time
+        wmhd = interp1(efit_time, wmhd, times)
+        tau_rad = wmhd / p_rad
         output = {
             "p_rad": p_rad,
             "dprad_dt": dprad,
@@ -607,12 +620,21 @@ class CmodPhysicsMethods:
             "p_icrf": p_icrf,
             "p_input": p_input,
             "radiated_fraction": rad_fraction,
+            "tau_rad": tau_rad,
         }
         return output
 
     @staticmethod
     @physics_method(
-        columns=["p_rad", "dprad_dt", "p_lh", "p_icrf", "p_input", "radiated_fraction"],
+        columns=[
+            "p_rad",
+            "dprad_dt",
+            "p_lh",
+            "p_icrf",
+            "p_input",
+            "radiated_fraction",
+            "tau_rad",
+        ],
         tokamak=Tokamak.CMOD,
     )
     def get_power(params: PhysicsMethodParams):
@@ -644,7 +666,12 @@ class CmodPhysicsMethods:
             p_oh = CmodPhysicsMethods.get_ohmic_parameters(params=params)["p_oh"]
         except mdsExceptions.TreeNODATA:
             p_oh = np.full(len(params.times), np.nan)
-        output = CmodPhysicsMethods._get_power(params.times, *values, p_oh)
+        wmhd, efit_time = params.mds_conn.get_data_with_dims(
+            r"\efit_aeqdsk:wplasm", tree_name="_efit_tree", astype="float64"
+        )  # [J], [s]
+        output = CmodPhysicsMethods._get_power(
+            params.times, *values, p_oh, wmhd, efit_time
+        )
         return output
 
     @staticmethod
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -619,6 +619,7 @@ class CmodPhysicsMethods:
         wmhd = interp1(efit_time, wmhd, times)
         with np.errstate(divide="ignore", invalid="ignore"):
             tau_rad = wmhd / p_rad
+        tau_rad[tau_rad == np.inf] = np.nan
         output = {
             "p_rad": p_rad,
             "dprad_dt": dprad,
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -514,7 +514,7 @@ class CmodPhysicsMethods:
                 r"\top.mflux:v0", tree_name="analysis"
             )  # [V], [s]
         except mdsExceptions.TreeException:
-            params.logger.warning(
+            params.logger.verbose(
                 r"v_loop: Failed to get \top.mflux:v0 data. Use \efit_aeqdsk:vloopt instead."
             )
             v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -652,7 +652,7 @@ class CmodPhysicsMethods:
         tokamak=Tokamak.CMOD,
     )
     def get_power(params: PhysicsMethodParams):
-        """
+        r"""
         NOTE: the timebase for the LH power signal does not extend over the full
             time span of the discharge. Therefore, when interpolating the LH power
             signal onto the "timebase" array, the LH signal has to be extrapolated
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -509,9 +509,14 @@ class CmodPhysicsMethods:
             A dictionary containing the calculated ohmic parameters, including
             "p_oh" and "v_loop".
         """
-        v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
-            r"\top.mflux:v0", tree_name="analysis"
-        )  # [V], [s]
+        try:
+            v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
+                r"\top.mflux:v0", tree_name="analysis"
+            )  # [V], [s]
+        except mdsExceptions.TreeException:
+            v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
+                r"\efit_aeqdsk:vloopt", tree_name="_efit_tree"
+            )  # [V], [s]
         if len(v_loop_time) <= 1:
             raise CalculationError("No data for v_loop_time")
 
@@ -665,7 +670,9 @@ class CmodPhysicsMethods:
                 kwa[p], kwa[t] = None, None
         # Ohmic power
         try:
-            kwa["p_ohm"] = CmodPhysicsMethods.get_ohmic_parameters(params=params)["p_oh"]
+            kwa["p_ohm"] = CmodPhysicsMethods.get_ohmic_parameters(params=params)[
+                "p_oh"
+            ]
         except mdsExceptions.TreeException:
             kwa["p_ohm"] = np.full(len(params.times), np.nan)
         # Plasma magnetic energy, and respective time base
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -648,31 +648,31 @@ class CmodPhysicsMethods:
             (Not-a-Number) values for times outside the LH timebase, and the NaN's
             will propagate into p_input and rad_fraction, which is not desirable.
         """
-        values = [
-            None
-        ] * 6  # List to store the time and values of the LH power, icrf power, and radiated power
+        # LH power, ICRF power, radiated power, and respective time bases
+        values = ["lh", "icrf", "rad"]
         trees = ["LH", "RF", "spectroscopy"]
-        nodes = [r"\LH::TOP.RESULTS:NETPOW", r"\rf::rf_power_net", r"\twopi_diode"]
+        nodes = [r"\TOP.RESULTS:NETPOW", r"\rf_power_net", r"\twopi_diode"]
+        kwa = {}
         # TODO: Find these nodes and add units
-        for i in range(3):
+        for val, tree, node in zip(values, trees, nodes):
+            p = f"p_{val}"
+            t = f"t_{val}"
             try:
-                sig, sig_time = params.mds_conn.get_data_with_dims(
-                    nodes[i], tree_name=trees[i], astype=None
+                kwa[p], kwa[t] = params.mds_conn.get_data_with_dims(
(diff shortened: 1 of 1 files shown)


disruption_py/machine/cmod/physics.py
```diff
@@ -507,19 +507,22 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
+                raise CalculationError("No data for v_loop_time")
+
+            li, efittime = params.mds_conn.get_data_with_dims(
+                r"\efit_aeqdsk:ali", tree_name="_efit_tree", astype="float64"
+            )  # [dimensionless], [s]
+            ip_parameters = CmodPhysicsMethods.get_ip_parameters(params=params)
+            r0 = params.mds_conn.get_data(
+                r"\efit_aeqdsk:rmagx/100", tree_name="_efit_tree"
+            )  # [m]
+        except mdsExceptions.TreeNODATA:
```
gtrevisan, line 524: is this needed?
didn't the previous version of the code return NaNs, as well, if TreeNODATA was triggered?

I might be wrong, of course. :)

disruption_py/machine/cmod/physics.py
```diff
@@ -589,8 +592,12 @@ def _get_power(times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm):
             # shots, excluding times for
             # which p_rad (uncalibrated) <= 1.e5 W
             dprad = np.gradient(p_rad, t_rad)
-            p_rad = interp1(t_rad, p_rad, times)
+            p_rad = interp1(t_rad, p_rad, times, fill_value=0)
             dprad = interp1(t_rad, dprad, times)
+            
+        if len(p_ohm) == 1 and np.isnan(p_ohm[0]):
```
gtrevisan, line 598: `if all(np.isnan(p_ohm))`?

disruption_py/machine/cmod/physics.py
gtrevisan: it should be caught here, and should effectively return a full-NaN array:
https://github.com/MIT-PSFC/disruption-py/blob/35b3f99b77552413216bdb762237c021350f18a0/disruption_py/core/physics_method/runner.py#L219-L228

yumouwei: So the problem is `get_power` calls `get_ohmic_parameters`; if `get_power` fails then it causes `get_ohmic_parameters` to fail and return nans for its outputs even if there are data for those signals in the tree.

https://github.com/MIT-PSFC/disruption-py/blob/efc63b986e36ef2e555b2f3df632ffe80a67fe66/disruption_py/machine/cmod/physics.py#L630-L641

I forgot why I put that try-except block in `get_power`. I guess the better way is to put it around line 639.

gtrevisan: who calls who? maybe you mistyped -- it's the caller that should catch the mdsplus exception, then.

yumouwei: Yeah there was a mistake. `get_power` calls `get_ohmic_parameters` and `_get_power`

gtrevisan: yep, okay, so then your try/except definitely needs to wrap line 639. :+1:

gtrevisan: (also, removing such indentation might suppress the current conflict with `dev`.)

disruption_py/machine/cmod/physics.py
```diff
@@ -591,8 +591,12 @@ def _get_power(times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm):
             # shots, excluding times for
             # which p_rad (uncalibrated) <= 1.e5 W
             dprad = np.gradient(p_rad, t_rad)
-            p_rad = interp1(t_rad, p_rad, times)
+            p_rad = interp1(t_rad, p_rad, times, fill_value=0)
             dprad = interp1(t_rad, dprad, times)
+
+        if all(np.isnan(p_ohm)):
```
gtrevisan, line 597: further thought -- do we want to set values to zero only for a full-NaN array?
could we have some values and some NaNs, instead?
another way, maybe more functional, is using `np.nan_to_num` which replaces NaN with 0 by default, wherever they might appear.

yumouwei: Actually that's a good idea, although I haven't found a shot where vloop has a shorter time base than efit so we won't be able to test this.

gtrevisan:
@yumouwei please double check my kwarg commit! also, was there an error about `p_oh` vs `p_ohm`?
is `_get_ohmic_parameters` supposed to return `p_oh` without a trailing M?

yumouwei:
Yes that's one of the annoying things with the CMOD branch. The Ohmic power is referred to as `p_ohm` everywhere else except in the column name which is called `p_oh`. This whole mess is needed to be consistent with the SQL table. To make it even worse, the DIII-D table has `p_ohm` while the EAST table has `p_oh` as their respective column name.

yumouwei:
`get_ohmic_parameters` will now use `\efit_aeqdsk:vloopt` when `\top:mflux:v0` data isn't available. This significantly changes the values of `p_input` and `radiated_fraction`.

(post change)
image

disruption_py/machine/cmod/physics.py
```diff
@@ -665,7 +670,9 @@ def get_power(params: PhysicsMethodParams):
                 kwa[p], kwa[t] = None, None
         # Ohmic power
         try:
-            kwa["p_ohm"] = CmodPhysicsMethods.get_ohmic_parameters(params=params)["p_oh"]
+            kwa["p_ohm"] = CmodPhysicsMethods.get_ohmic_parameters(params=params)[
```
gtrevisan, line 673: well at this point you can break it up with `output =` or `result =` and then assign it into `kwa` on the following line.

disruption_py/machine/cmod/physics.py
```diff
@@ -509,9 +509,14 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
             "p_oh" and "v_loop".
         """
-        v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
-            r"\top.mflux:v0", tree_name="analysis"
-        )  # [V], [s]
+        try:
+            v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
+                r"\top.mflux:v0", tree_name="analysis"
+            )  # [V], [s]
+        except mdsExceptions.TreeException:
```
gtrevisan, line 516: maybe we should add a debug* statement here to track this fallback.
*: it will be a verbose statement once we merge that PR.

disruption_py/machine/cmod/physics.py
```diff
@@ -559,20 +566,26 @@ def _get_power(times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm):
             The time array corresponding to radiated power.
         p_ohm : np.ndarray
             The ohmic heating power.
+        wmhd : np.ndarray
+            The total plasma energy.
+        efit_time : np.ndarray
+            The EFIT time base
```
gtrevisan, line 572: missing final full stop for consistency.

disruption_py/machine/cmod/physics.py
```diff
@@ -559,20 +566,26 @@ def _get_power(times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm):
...
 
         Returns
         -------
         dict
             A dictionary containing the calculated power values, including
-            "p_rad", "dprad_dt", "p_lh", "p_icrf", "p_input", and "radiated_fraction".
+            "p_rad", "dprad_dt", "p_lh", "p_icrf", "p_input", "radiated_fraction", and "tau_rad".
+
+        Last major update by William Wei on 1/6/2025
```
gtrevisan, line 580: I am now unsure whether to keep adding changelog lines like this one in the docstrings, since we (mostly you) touched up most of the physics parameters anyways... I welcome suggestions on this.

disruption_py/machine/cmod/physics.py
```diff
@@ -591,24 +604,43 @@ def _get_power(times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm):
             # shots, excluding times for
             # which p_rad (uncalibrated) <= 1.e5 W
             dprad = np.gradient(p_rad, t_rad)
-            p_rad = interp1(t_rad, p_rad, times)
+            p_rad = interp1(t_rad, p_rad, times, fill_value=0)
             dprad = interp1(t_rad, dprad, times)
+
+        # Replace nans with zeros in p_ohm
+        p_ohm = np.nan_to_num(p_ohm, nan=0.0)
```
gtrevisan, line 611: have you tried replacing it in-place, as well?
```python
np.nan_to_num(p_ohm, copy=False, nan=0.0)
```

disruption_py/machine/cmod/physics.py
```diff
@@ -621,23 +653,33 @@ def get_power(params: PhysicsMethodParams):
...
-            None
-        ] * 6  # List to store the time and values of the LH power, icrf power, and radiated power
+        # LH power, ICRF power, radiated power, and respective time bases
+        values = ["lh", "icrf", "rad"]
         trees = ["LH", "RF", "spectroscopy"]
-        nodes = [r"\LH::TOP.RESULTS:NETPOW", r"\rf::rf_power_net", r"\twopi_diode"]
+        nodes = [r"\TOP.RESULTS:NETPOW", r"\rf_power_net", r"\twopi_diode"]
+        kwa = {}
         # TODO: Find these nodes and add units
-        for i in range(3):
```
gtrevisan, line 630: would you be able to take care of the TODO here above? since we are at it? thanks!

disruption_py/machine/cmod/physics.py
```diff
@@ -621,23 +653,33 @@ def get_power(params: PhysicsMethodParams):
...
+        for val, tree, node in zip(values, trees, nodes):
+            p = f"p_{val}"
+            t = f"t_{val}"
             try:
-                sig, sig_time = params.mds_conn.get_data_with_dims(
-                    nodes[i], tree_name=trees[i], astype=None
+                kwa[p], kwa[t] = params.mds_conn.get_data_with_dims(
```
gtrevisan, line 666: does this make sense to you? it's better than the `*values` trick but I'm not happy with how this turned out following my commit. any improvement is welcome.

gtrevisan (changes requested):

disruption_py/machine/cmod/physics.py
```diff
@@ -591,24 +604,44 @@ def _get_power(times, p_lh, t_lh, p_icrf, t_icrf, p_rad, t_rad, p_ohm):
...
+        # Calculate radiated fraction
         p_input = p_ohm + p_lh + p_icrf
         rad_fraction = p_rad / p_input
         rad_fraction[rad_fraction == np.inf] = np.nan
+
+        # Calculate radiation confinement time
+        wmhd = interp1(efit_time, wmhd, times)
+        with np.errstate(divide="ignore", invalid="ignore"):
+            tau_rad = wmhd / p_rad
+        tau_rad[tau_rad == np.inf] = np.nan
```
amdecker, line 622: Minor: instead of 
```python
with np.errstate(divide="ignore", invalid="ignore"):
    tau_rad = wmhd / p_rad
tau_rad[tau_rad == np.inf] = np.nan
```
we can do
```python
tau_rad = np.full_like(wmhd, np.nan)
np.divide(wmhd, p_rad, out=tau_rad, where=p_rad != 0)
```
to more explicitly handle divisions by zero. This ensures we only perform the division where `p_rad` is non-zero and anywhere we would have divided by is a nan instead of inf. I like this better than ignoring the errors because it is more explicit, but lmk if you prefer the `with` statement

gtrevisan: no offense, but that's ugly! :chipmunk: 
what about something more physics-based, but still explicit, like so?
```python
tau_rad = wmhd / np.where(p_rad != 0, p_rad, np.nan)
```
that should be sufficient by itself. LMK!

disruption_py/machine/cmod/physics.py
gtrevisan: @amdecker thoughts on making this prettier yet robust and explicit?

disruption_py/machine/cmod/physics.py
gtrevisan: you can now use the verbose level!

disruption_py/machine/cmod/physics.py
gtrevisan: I thought about it -- we should put here only actionable intel, that is, references to PRs that modified the physics or logic.
in such a way we can quickly backtrack through physics changes without being submerged by github blame info that would also include lots of refactoring/linting/logging-based PRs.
so something like:
```python
"""
References
----------
- https://github.com/MIT-PSFC/disruption-py/pull/367
"""
```

or any other format appropriate for reading _and_ for mkdocs-generated documentation.
@amdecker?

yumouwei: Added References block in docstring.

disruption_py/machine/cmod/physics.py
```diff
@@ -676,7 +676,7 @@ def get_power(params: PhysicsMethodParams):
             t = f"t_{val}"
             try:
                 kwa[p], kwa[t] = params.mds_conn.get_data_with_dims(
-                    node, tree_name=tree, astype=None
+                    node, tree_name=tree
```
gtrevisan, line 679: bug I introduced inadvertently in:
https://github.com/MIT-PSFC/disruption-py/pull/364/files#diff-13362dcb51f88ed4cb8a9a082dca2ce1997dc3e8c9e3c2635e9e565b64438120

disruption_py/machine/cmod/physics.py
gtrevisan: thanks, we will have to remove all lines like "Last major update" by replacing them with PR links, but it can be a separate PR.

disruption_py/machine/cmod/physics.py
```diff
@@ -662,9 +662,9 @@ def get_power(params: PhysicsMethodParams):
             will propagate into p_input and rad_fraction, which is not desirable.
 
         LH, ICRF, & radiated power data sources:
-        - lh: \lh::top.results.netpow [kW]
-        - icrf: \rf::top.antenna.results.pwr_net_tot [MW]
-        - rad: \spectroscopy::top.bolometer.twopi_diode [kW]
+        - LH: \lh::top.results.netpow [kW]
+        - ICRF: \rf::top.antenna.results.pwr_net_tot [MW]
+        - Radiated: \spectroscopy::top.bolometer.twopi_diode [kW]
```
gtrevisan, line 667: what error did you get? it worked on my end without this change. what version of pylint?

yumouwei: I added the `r"""` before the docstring which you introduced in your commit as well.

gtrevisan: you reverted the wrong commit! lol.

disruption_py/machine/cmod/physics.py
gtrevisan: you also added it to the wrong method (private vs public) but that's okay for now!
I opened up another issue in order to track this.

disruption_py/machine/cmod/physics.py
```diff
@@ -509,9 +509,17 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
         """
-        v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
-            r"\top.mflux:v0", tree_name="analysis"
-        )  # [V], [s]
+        try:
+            v_loop, v_loop_time = params.mds_conn.get_data_with_dims(
+                r"\top.mflux:v0", tree_name="analysis"
+            )  # [V], [s]
+        except mdsExceptions.TreeException:
+            params.logger.warning(
```
gtrevisan, line 517: should we consider this a warning or just a fallback, thus using verbose?

gtrevisan (approved):
LGTM, please have Alex test this out, then I'll merge!

AlexSaperstein (approved):
Looks good to me. I compared tau_rad for shot 1120501008 between my onw_dev code calc and this update and they agree.

Issue #346, opened 2024-11-14
Add option to not run a specific method


AlexSaperstein:
when specifying the RetrievalSettings, it would be useful to be able to specify specific methods to not be run. the advantage of this would be that you can still specify "all" for the run tags, and then specifically pick the methods (or even columns) you don't want run, which may be faster and easier to do if there are only some methods you don't want run.

This came up when trying to troubleshoot some issues pulling data that appears to be related to the get_peaking_factors and get_ts_parameters methods, and we would like to be able to just shut off these methods only.

gtrevisan:
hmm, I think at this stage you should rather get a list of all methods and drop what you do not want.

@amdecker, could you please share a snippet to accomplish this?
if this is not achievable easily, then we can try and implement it in a user-friendly way.

amdecker:
Here's a small example showing how to get physics methods for all but a few specified physics methods.
```python
from disruption_py.core.physics_method.runner import get_all_physics_methods
from disruption_py.settings.retrieval_settings import RetrievalSettings
from disruption_py.workflow import get_shots_data

from disruption_py.machine.cmod import METHOD_HOLDERS

all_physics_methods = get_all_physics_methods(METHOD_HOLDERS)
method_names = [method.__name__ for method in all_physics_methods]

# E.g. retrieve data for everything except get_peaking_factors and get_ts_parameters
method_names.remove("get_peaking_factors")
method_names.remove("get_ts_parameters")

retrieval_settings = RetrievalSettings(
    time_setting="efit",
    run_tags=[],
    run_methods=method_names,
)
shot_data = get_shots_data(
    tokamak="cmod",
    shotlist_setting=[1150805012, 1150805013, 1150805014],
    retrieval_settings=retrieval_settings,
    num_processes=1,
)
```

gtrevisan:
okay, so roughly the only required _addition_ is something like this -- I didn't test it, I just wanted to see the impact of this new block into an existing script:
```python
# exclude some methods
from disruption_py.core.physics_method.runner import get_all_physics_methods as gapm
from disruption_py.machine.cmod import METHOD_HOLDERS as MH
exclude = ["get_peaking_factors", "get_ts_parameters"]
methods = [m.__name__ for m in gapm(MH) if m not in exclude]
```
let us know if this works for your purpose!

AlexSaperstein:
It does!

I checked pulling data on dev for shot 1051123016 using `run_tags=['all']` and got a crash

[...]

PR #368, opened 2024-12-27
Allow quick exclusion from run_methods

lajz:
It seems as though there may be a desire to not include certain methods from a tag, this PR introduces the ability to prefix method names with a tilda as a means of specifying that is should NOT be run.


gtrevisan (approved):
thanks for the PR, Josh!

we tried to massage the current system of specifying methods to run:
- #346 

with a PR that makes it trivial to fetch the method names so one can drop the undesired ones:
- #369

I'm not in love with the way the user has to specify methods at present -- eg the default run_tag `all` has tricked many into thinking they were running only the specified methods. :smile:

on one hand, I think at a certain point we should revamp the full user-facing logic to be clearer and more intuitive...
... but on the other hand, this PR is fine with me. it slightly worsens the readability of the if/and/or logic in that section but it does offer another quick way to exclude methods.

any thoughts, @amdecker?

if any of you sees a quick way to revamp this inclusion/exclusion logic, we should be able to implement it fairly quickly.
my current thoughts would be to drop the tags (currently unused AFAIK, we moved the experimental ones to the drafts folder), and just have a specific `__all__`  string to run all methods, so that these two examples:
```python
# all
do.run(run_tags=["all"], run_methods=[])
# some
do.run(run_tags=[], run_methods=["method1"])
```
might become:
```python
# all
do.run(run_methods=["__all__"])
# some
do.run(run_methods=["method1"])
```
but I haven't given this much thought, so there might be better ways.
I kind of like the pythonic way of expressing negatives with the tilde, though!

lajz:
I think having a tags.py or tags.yaml file for defining tags may be more intuitive, thoughts?

gtrevisan:
well, at this time we did not get any usage for tags -- as I mentioned, experimental stuff has been discontinued into the drafts folder... and in any case, it should be trivial to re-add it later on.

most users think in terms of what columns they want, or what methods.
in Alex's case, he wanted to exclude a method that was yielding a lot of errors.

I like my earlier example because it minimizes the number of arguments to convey the user's intent, and is straightforward to document.
we could also code it up so that an empty `run_methods=[]` argument means to run them _all_, while any content means to run only the specified methods.

amdecker:
What do you guys think of getting rid of tags -- I agree the behavior with `run_tags=["all"]` as default is confusing -- but applying this cool Pythonic tilde idea to the `run_methods` parameter. Example usage
```python
do.run(run_methods=["method1"]) # runs only method1
do.run(run_methods=["~method1"]) # runs all except method1
do.run() # runs all

```

gtrevisan:
yep, that's exactly what I proposed! I agree. let's do that.
@amdecker do you have bandwidth to code it up and supersede both of these PRs?

lajz:
beautiful

disruption_py/core/physics_method/runner.py
```diff
@@ -171,15 +172,21 @@ def filter_methods_to_run(
...
-        if tags is not None and bool(
-            set(bound_method_metadata.tags).intersection(tags)
-        ):
-            methods_to_run.append(bound_method_metadata)
-        elif methods is not None and bound_method_metadata.name in methods:
-            methods_to_run.append(bound_method_metadata)
-        elif columns is not None and bool(
-            set(bound_method_metadata.columns).intersection(columns)
-        ):
+        should_run = (
```
gtrevisan, line 175: IDK, for maintainability we might want to break this up more.

gtrevisan (changes requested):
AFAIU, this breaks the original intuition, that is:
```python
run_methods = ["~get_peaking_factors"]
```
is opposed to run everything BUT the specified method.
maybe something like:
```python
methods is not None and all(m.startswith("~") for m in methods)
```
to be added to `should_run`?

amdecker:
Good catch! Cleaned up the logic a bit too

gtrevisan (approved):
LGTM after a quick test, but I definitely did not investigate all the edge cases.
tagging Alex and William so they can test this out a bit further.

on a separate note:
if I run `get_ohmic_parameters` for C-MOD, which does include running `get_ip_parameters`, shouldn't I get a notice that the latter has run, as well?

amdecker:
> if I run get_ohmic_parameters for C-MOD, which does include running get_ip_parameters, shouldn't I get a notice that the latter has run, as well?

We only log the collected physics methods in `runner.py`, so since `get_ip_parameters` is called from within `get_ohmic_parameters` it won't be logged. If we want to log each call to a physics method, we could easily add a decorator or modify the `@physics_method` decorator to log whenever it is called. LMK if that's something you would like

gtrevisan:
yes, that's definitely the desired behavior! feel free to put it on the back burner as a low-priority improvement.
(AFAIK we should have just _one_ call for each method due to method caching anyways.)
waiting for further tests, then merging!

AlexSaperstein:
I tried testing this out. most of it seems fine, but it seems real odd that in order to just return all columns you have to set `run_methods = None`. In my opinion having `run_methods = ['all']` feels a little more intuitive if that could work.

gtrevisan:
you can just _not set it_, though. eg `get_shots_data()` :wink:

yumouwei:
`examples/defaults.py` does have `run_methods=[], run_columns=[]` as its default setting. A user should know how to use the actual default setting (i.e. run everything) without digging into the docstring of `RetrievalSettings`.

gtrevisan (changes requested):
@amdecker could you do another pass and make sure the docs are clear?
runs methods should default to None, rather than empty list, and the defaults example should mimic that. thanks!

amdecker:
>examples/defaults.py does have run_methods=[], run_columns=[] as its default setting

Thanks for pointing that out! That was a mistake, and I have now updated `defaults.py` to have `run_methods=None, run_columns=None`. 

>But then maybe the default example could make that option clearer.

I added a comment in `defaults.py` to help explain, and here's what the `RetrievalSettings` docs look like for `run_methods` and `run_columns`:
image

LMK if you have suggestions for how to make the options clearer!

examples/defaults.py
```diff
@@ -10,18 +10,17 @@
...
     # data settings
     cache_setting=None,
-    efit_nickname_setting="disruption",
-    # method selection
-    run_methods=[],
-    run_tags=["all"],
-    run_columns=[],
-    only_requested_columns=False,
+    efit_nickname_setting="disruption",  # defaults to use "disruption" tree for disruptive shots
+    # method selection.
```
gtrevisan, line 14: @yumouwei
I don't think these comment changes are useful.
please avoid inline comments, but feel free to add some _additional_ logic in the "header comments", eg after
```python
# method selection
```

examples/defaults.py
```diff
@@ -10,18 +10,17 @@
...
+    only_requested_columns=False,  # defaults to return all columns retrieved by the methods
     custom_physics_methods=[],
     # timebase settings
-    time_setting="disruption_warning",
-    domain_setting="full",
+    time_setting="disruption_warning",  # defaults to return data in "disruption_warning" time base
+    domain_setting="full",  # defaults to get entire shot
     use_cache_setting_timebase=False,
-    interpolation_method="linear",
+    interpolation_method="linear",  # defaults to use linear interpolation
```
gtrevisan, line 23: I stress additional because arguably this here is a tautology!

gtrevisan (changes requested):

yumouwei (approved):

gtrevisan (approved):

yumouwei (approved):

PR #370, opened 2025-01-02
Fix logs for MDSplus tree opening

amdecker:
## Problem
The trace logs are cluttered with redundant opening tree logs because previously we logged whenever the `open_tree` method of `MDSConnection` was called, even when we did not open a new connection.

## Proposed solution
Clean up the logs by only logging when we open a new tree with `self.conn.openTree(tree_name, self.shot_id)`. This also means we now log the MDS tree name rather than a nickname.

Previously:
image

Now:
image

## Unchanged
We also have multiple database initialization logs even when there is only one shot. We pass a database instance to the `ShotlistSettingParams` in case someone wants to use an SQL query to get the shotlist. 

```
07:35:16.301 [ DEBUG ] Database initialization: decker@alcdb2/logbook
07:35:16.334 [ INFO  ] Starting workflow: 1 shot / 1 process
07:35:16.343 [ DEBUG ] Database initialization: decker@alcdb2/logbook     
```
The database initializer is lazy, so when we initialize the database we do not create a new connection. We log new connections separately.

### DIFF
--- a/disruption_py/inout/mds.py
+++ b/disruption_py/inout/mds.py
@@ -108,7 +108,6 @@ class MDSConnection:
         If the specified tree_name is a nickname for a tree_name, will open the tree
         that it is a nickname for.
         """
-        logger.trace("Opening tree: {tree_name}", tree_name=tree_name)
         if (
             tree_name not in self.tree_nicknames
             and tree_name in self.tree_nickname_funcs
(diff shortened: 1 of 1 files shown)


gtrevisan (approved):
much better now for MDSplus calls! :+1: 

I still think we might improve the DB init, but we'll chat about it next time.

PR #378, opened 2025-01-09
Refactor polarity

amdecker:
## Problem: 
Workflow failed with duplicate pylint error https://github.com/MIT-PSFC/disruption-py/actions/runs/12579637452/job/35060151149

## Proposed solution
Tweaking the code (https://github.com/MIT-PSFC/disruption-py/pull/363/commits/cd38156c225f2c37728551fb7d95b22c6045b7ad) to avoid the error is a temporary fix. This PR refactors the block handling polarity into a separate method so it can be reused without duplicate code.

### DIFF
--- a/disruption_py/machine/d3d/physics.py
+++ b/disruption_py/machine/d3d/physics.py
@@ -469,6 +469,34 @@ class D3DPhysicsMethods:
         return {"n_e_rt": ne_rt, "greenwald_fraction_rt": g_f_rt, "dn_dt_rt": dne_dt_rt}
 
     @staticmethod
+    def get_polarity(params: PhysicsMethodParams):
+        """
+        Get the polarity of the plasma current. Returns the first value if the polarity
+        is not constant.
+
+        Parameters
+        ----------
+        params : PhysicsMethodParams
+            Parameters containing MDS connection and shot information.
+
+        Returns
+        -------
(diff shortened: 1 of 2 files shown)


gtrevisan (commented):
I fully agree that this is the proper fix, and thanks for digging around and discovering there were _three_ repetitions, not just two!

I'm not sure about the current status of the new function, though.
could we name it privately to signify it's not supposed to be run by the user?
in this way (both being public and being named as a getter like other physics methods) I think one could easily get confused.
- what about naming it privately? would it hinder usage from `domain_setting.py`? would it require a pylint exception?
- what about changing its name from `get_` to something else?

in a way, and AFAIK, so far we've kind of reserved the "getter status" to physics methods.
thoughts?

gtrevisan:
what about moving such utility method in a new container class eg `CmodUtilMethods`?

gtrevisan:
> what about moving such utility method in a new container class eg `CmodUtilMethods`?

@amdecker was a there a reason for not creating a new class? what if we had to create other utility methods?

gtrevisan (changes requested):
also, if you can, please fix conflicts, otherwise I'll get to it tomorrow! thanks.

gtrevisan (approved):

PR #381, opened 2025-01-14
Cycle cache example

amdecker:
Simple example to show how to use the previous output of a `get_shots_data` request as the cache for a new `get_shots_data` request.


examples/cycle_cache_example.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+
+"""
+Example usage of `get_shots_data` when providing cache data from a previous run.
+"""
+
+from disruption_py.settings.retrieval_settings import RetrievalSettings
+from disruption_py.workflow import get_shots_data
+
+shots = [1150805012]
```
gtrevisan, line 10: could we use `get_tokamak_test_shotlist`, instead?

examples/cycle_cache_example.py
```diff
@@ -0,0 +1,22 @@
...
+cached_shot_data = get_shots_data(
+    tokamak="cmod",
```
gtrevisan, line 12: could we make it tokamak agnostic?

examples/cycle_cache_example.py
```diff
@@ -0,0 +1,22 @@
...
+    shotlist_setting=shots,
+)
+
+shot_data = get_shots_data(
+    tokamak="cmod",
+    shotlist_setting=shots,
+    retrieval_settings=RetrievalSettings(
```
gtrevisan, line 19: could we assign retrieval_settings before the get_shots_data call as usual to make it even clearer?

gtrevisan (changes requested):
thanks! I think this will be useful to lots of people! :rocket:

here follow my current thoughts on this feature showcase:

0. the user (that is, our new script) creates a long-lived cache for this, either csv but maybe hdf5? eg `~/.cache/disruption-py/my-workflow.h5`
1. first run that gets data for half of the test shotlist
   - output setting points at that cache + df in memory
2. second identical run to show that no time gets wasted.
   - cache setting points to df in memory.
4. third run of full test shotlist to show that only missing data gets pulled, rest is cached.
    - cache setting points to hdf5 on disk.

EDIT: I'd rather call this file `hdf_cache.py` or `csv_cache.py` for brevity.

what do you think?

disruption_py/settings/cache_setting.py
```diff
@@ -151,12 +151,57 @@ def _get_cache_data(self, params: CacheSettingParams) -> pd.DataFrame:
         return self.cache_data
 
 
+class HDF5CacheSetting(CacheSetting):
```
gtrevisan, line 154: huh! I was sure this was already doable.
isn't this included (or include-able) in the `DFCacheSetting`, since those are dataframes anyway?

disruption_py/settings/output_setting.py
```diff
@@ -449,7 +449,7 @@ def _output_shot(self, params: OutputSettingParams):
 
         output_result.to_hdf(
             self.filepath,
-            key=f"df_{params.shot_id}",
+            key=str(params.shot_id),
```
gtrevisan, line 452: maybe this is why? lol.

```
[WARNING] disruption-py-z-VhksUo-py3.10/lib/python3.10/site-packages/tables/path.py:137: NaturalNameWarning: object name is not a valid Python identifier: '1150805016'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
```

gtrevisan (commented):
that's great!

disruption_py/settings/cache_setting.py
amdecker: We do use pandas to read both into a DataFrame, but the DFCacheSetting is for passing an actual DataFrame as cache_setting, whereas the new HDF5CacheSetting and CSVCacheSetting are for passing a file path.

gtrevisan (approved):

PR #386, opened 2025-01-21
Add initial framework support for EAST

gtrevisan:
- allow different DB dialect,
- allow fuzzy driver specification,
- add version getter for DB,
- add EAST minimal config for MDSplus and SQL.


disruption_py/machine/cmod/config.toml
```diff
@@ -2,7 +2,7 @@
 mdsplus_connection_string = "alcdata-archives"
 
 [cmod.inout.sql]
-driver = "ODBC Driver 18 for SQL Server"
+driver = "ODBC"
```
amdecker, line 5: If we have the exact driver is there a reason we should use the short name & rely on 'fuzzy' resolution?

gtrevisan: I saw differences between the different machines wrt the version number, so it seemed to me this was the easiest check of what driver one wants to use.
in any case, it should be clear enough that it could be adapted at will when the circumstances fail.

to some extent, I currently see MFE WSs using ODBC, omega using FreeTDS, and EAST machines using MySQL...

amdecker (approved):

Issue #379, opened 2025-01-09
Fix references for physics methods


gtrevisan:
in order to easily provide a physics-based reference for all our physics methods, we should provide a `References` section in the docstrings which link to the relevant physics-based PRs or external reading resources.

watch out! in order to be properly parsed by mkdocs, these references should be in the public physics method, not in its private counterpart.
this will make very clear who carried out which physics improvement and when, while the classic git blame approach would highlight countless code-based linting, logging, and refactoring changes.

we should first figure out the best format for mkdocs support: https://mkdocstrings.github.io/python/

eg: https://mit-psfc.github.io/disruption-py/usage/physics_methods/d3d_built_in_method_reference/#disruption_py.machine.d3d.physics.D3DPhysicsMethods.get_density_parameters

to test out mkdocs locally:
```
poetry install --with docs
poetry run mkdocs serve
```

gtrevisan:
so for that one, probably something like an unordered list:
```
"""
Get electron density from EFIT, then compute dn_dt and Greenwald_fraction.

References
----------
https://github.com/MIT-PSFC/disruption-py/blob/matlab/DIII-D/get_density_parameters.m
https://github.com/MIT-PSFC/disruption-py/issues/238
https://github.com/MIT-PSFC/disruption-py/pull/249
"""
```

at worst, we might have to start putting some metadata in the PRs themselves.

gtrevisan:
or maybe even more succinct eg:
```
"""
Get electron density from EFIT, then compute dn_dt and Greenwald_fraction.

References
----------
- original source: [get_density_parameters.m](https://github.com/MIT-PSFC/disruption-py/blob/matlab/DIII-D/get_density_parameters.m)
- issues: #[238](https://github.com/MIT-PSFC/disruption-py/issues/238)
- pull requests: #[249](https://github.com/MIT-PSFC/disruption-py/pull/249)
"""
```
which yields:
![image](https://github.com/user-attachments/assets/d6397614-dbbe-4fc9-ac2e-50148416f9e9)

amdecker:
I like the references section! 

Another idea I had would be to use footnotes
![Image](https://github.com/user-attachments/assets/b10964fc-51a7-4f06-962d-921fc00f7403)

```
Calculate the normalized beta
...
beta_n is defined as the total plasma beta (p/(B^2/2mu_0))
divided by Ip [MA]/(a*Bt).[^1]
...
[...]

PR #390, opened 2025-01-28
LaTeX support for documentation

amdecker:
Adds [KaTeX](https://squidfunk.github.io/mkdocs-material/reference/math/#katex-mkdocsyml) support for MkDocs.

image

We can do inline LaTeX with `$...$` and blocks with `$$...$$`

I changed two physics methods to use LaTeX. I will wait on footnote & reference improvements to do more.
- https://github.com/MIT-PSFC/disruption-py/issues/379#issuecomment-2616616889


gtrevisan (approved):
LGTM! thanks!

(maybe you could add a newline to those two files that end with :no_entry_sign: ?)

yumouwei (approved):
It works and I like it, so let's merge it.

PR #391, opened 2025-01-28
Log physics method time in decorator

amdecker:
## Problem
As discussed in 
- https://github.com/MIT-PSFC/disruption-py/pull/368#issuecomment-2606837596

Some physics methods are called by others (e.g. `get_ohmic_parameters` calls `get_ip_parameters`) but the time taken for the inner calls are not logged.

## Proposed solution
Do not log in `runner.py` for each collected method, but instead in the `physics_method` decorator so all physics methods, regardless of how they are called, get logged.

## Verification
```
python examples/simple.py -m get_ohmic_parameters
```
Now logs both methods. 

image

### DIFF
--- a/disruption_py/core/physics_method/decorator.py
+++ b/disruption_py/core/physics_method/decorator.py
@@ -5,6 +5,7 @@ This module provides a decorator to signify methods that calculate physics quant
 """
 
 import time
+from functools import wraps
 from typing import Callable, List, Union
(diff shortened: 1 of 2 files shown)


gtrevisan (changes requested):
a couple of tests reliably fail on this branch, otherwise it LGTM.

PR #398, opened 2025-02-07
Re-sort all TOML files

gtrevisan:
- move toml-sort config into pyproject
- introduce two tiers of TOML checks
- deeper files (eg machine configs) sort _all_ keys


disruption_py/config.toml
```diff
@@ -1,13 +1,13 @@
 [default]
-time_const = 1e-6
 max_processes = 10
 max_shot_time = 7.0 # [s]
+time_const = 1e-6
 
 [default.inout.sql]
+protected_columns = ["shot", "time"]
```
gtrevisan, line 7: @amdecker is this going to create problems when testing, AFAYK?
I think this is much better -- independent of the fact that it's _also_ in alphabetical order, now.

amdecker: Should be ok!

amdecker (approved):

PR #400, opened 2025-02-11
Simplify lint jobs with matrix

gtrevisan:
- simplify lint jobs with matrix
- add toml-sort
- minor fix to yaml exclusions

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,13 +17,16 @@ on:
 jobs:
 
 
-  black:
+  lint:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


amdecker (approved):
How long are we keeping poetry around for? Would this be a good opportunity to switch linting over to `uv run ...`? Otherwise looks good!

gtrevisan:
I wanted first to add support for uv, so that I could play with it myself, but yes, I will definitely switch all CI/CD workflows to uv in the upcoming future.
I am now maintaining both poetry and uv for people to use on our workstations/clusters.

PR #401, opened 2025-02-18
Fix measurement of test times

gtrevisan:
add a simple wrapper test with the only purpose of triggering the pytest fixtures so that time spent eg collecting data will be assigned to this test.
as a consequence, test_data_columns -- or, more precisely, the first parameterized data column test -- should drop from being one of the slowest tests to being as fast as the others.

ideally, we should have similar wrappers in other test files in order to better diagnostic drastic slowdowns in test execution.

returning data to avoid the unused variable linter error.

### DIFF
--- a/tests/test_against_cache.py
+++ b/tests/test_against_cache.py
@@ -107,6 +107,18 @@ def cache_data_fixture(
     return cache_data
 
 
+def test_data_fetch(
+    shotlist: List[int],
+    fresh_data: Dict[int, pd.DataFrame],
+    cache_data: Dict[int, pd.DataFrame],
+    expected_failure_columns: List[str],
+):
+    """
+    Convenience test that triggers some fixtures to get more accurate time measurements.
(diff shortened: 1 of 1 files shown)


amdecker (commented):
Current behavior before this PR when running `pytest --durations=0` to print the slowest tests is to differentiate between `setup` times and `call` times for the first call of a test using fixtures.

e.g.
```
10.37s setup    tests/test_against_cache.py::test_data_columns[beta_n]
...
0.01s call     tests/test_against_cache.py::test_data_columns[beta_n]
```
I see that this PR adds a new test so all the setup time is devoted to a new test. Is it needed for the purpose of diagnosing test slowdown times given there is already a distinction between `setup` and `call`?

gtrevisan:
wow, I definitely miss that! yes, that was my intended purpose, I'll recheck the output more carefully and maybe close this. thanks!

Issue #383, opened 2025-01-14
Unreasonably large te_width for several CMOD shots


yumouwei:
# Statement of the problem

`te_width` outputs are unreasonably large for several shots.

List of shots I found in the `116` range:

> [1160615019, 1160615012, 1160615020, 1160615021, 1160615022, 1160616007, 1160923020, 1160923019, 1160929019, 1160930036]

# Example
(disruption_py==v0.8.0)

![image](https://github.com/user-attachments/assets/b7a0a823-fecb-4b1c-b17a-50b445ca4da6)

gtrevisan:
the physics method could definitely be made more robust by introducing a physically-meaningful threshold above which to discard any te_width, or maybe by checking the other byproducts of the gaussian fitting to assess the quality of the fit.
could you investigate a couple of those cases? thanks!

yumouwei:
Regarding setting an upper limit, one way I can think of is to use something based on `b_minor` (the TS chords are spaced vertically in the machine). Using the machine parameters listed [here](https://cmodwiki.psfc.mit.edu/index.php/Alcator_Parameters) we get `b_minor = a_minor * kappa ~< 0.4m` which should be sufficient as `te_width` (HWHM) is typically around 0.2m for a good shot.

gtrevisan:
closed with:
- #402

PR #402, opened 2025-02-18
Improve C-MOD Te width computation

yumouwei:
# Original problem
- #383 

# Analysis of the problem
- Several outer cords in TE_RZ appear to have spikes that shoot up to unreasonably large values. Jerry Hughes commented that these are likely caused by errors from the Thomson's hardware.
- To compensate these, Jerry recommends including the Te error (`TE_ERR`) in the gaussian fit. 

(1160615021)
image

# Implemented changes

- Include `TE_ERR` as the uncertainty of the Te data points in `curve_fit`. Doing this alone resolves all of the identified problem. 
  - 1150805022 time [24] which is a previously identified bad fit is now automatically rejected as the fit couldn't converge.
- Check and reject unreasonable values of the fitted gaussian mean and HWHM.

# Sample outputs

image
image

### DIFF
--- a/disruption_py/machine/cmod/physics.py
+++ b/disruption_py/machine/cmod/physics.py
@@ -992,7 +992,7 @@ class CmodPhysicsMethods:
         return output
 
     @staticmethod
-    def _get_ts_parameters(times, ts_data, ts_time, ts_z, z_sorted=False):
+    def _get_ts_parameters(times, ts_data, ts_time, ts_z, ts_error, z_sorted=False):
         """
         Calculate the Thomson scattering temperature width parameters.
 
@@ -1006,6 +1006,8 @@ class CmodPhysicsMethods:
             Corresponding time values for the temperature data.
         ts_z : array_like
             Vertical coordinate values corresponding to the temperature data.
+        ts_error: array_like
+            2D array of Thomson scattering temperature error data.
         z_sorted : bool, optional
             If True, assumes `ts_z` is already sorted. Default is False.
 
@@ -1019,6 +1021,7 @@ class CmodPhysicsMethods:
             idx = np.argsort(ts_z)
             ts_z = ts_z[idx]
             ts_data = ts_data[idx]
+            ts_error = ts_error[idx]
         # init output
         te_hwm = np.full(len(ts_time), np.nan)
         # select valid times
@@ -1029,6 +1032,7 @@ class CmodPhysicsMethods:
         for idx in valid_times:
             # select non-zero indices
             y = ts_data[:, idx]
+            y_error = ts_error[:, idx]
             (ok_indices,) = np.where(y != 0)
             # skip if not enough points
             if len(ok_indices) < 3:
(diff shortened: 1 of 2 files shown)


disruption_py/machine/cmod/physics.py
```diff
@@ -1031,28 +1034,33 @@ def _get_ts_parameters(times, ts_data, ts_time, ts_z, z_sorted=False):
...
             try:
-                _, _, psigma = gaussian_fit(z, y, guess)
+                _, pmean, psigma = gaussian_fit(z, y, guess, y_error, True)
             except RuntimeError as exc:
                 if str(exc).startswith("Optimal parameters not found"):
                     continue
                 raise exc
+            # rescale from sigma to HWHM
+            # https://en.wikipedia.org/wiki/Full_width_at_half_maximum
+            hwm = np.abs(psigma) * np.sqrt(2 * np.log(2))
```
gtrevisan, line 1058: you should rescale this once before returning, not in the for loop.

disruption_py/machine/cmod/physics.py
```diff
@@ -1031,28 +1034,33 @@ def _get_ts_parameters(times, ts_data, ts_time, ts_z, z_sorted=False):
...
+            # reject points with unphysical mean or hwm
+            if pmean < -0.35 or pmean > 0.35 or hwm > 0.35:
```
gtrevisan, line 1060: here in the for loop I'd just check `pmean`, eg:
```
if abs(pmean) < 0.35:
    # store output
    te_hwm[idx] = hwm
```

and then, after the loop, I'd do a single check on hwhm > 0.35 with a where.

disruption_py/machine/cmod/physics.py
```diff
@@ -1031,28 +1034,33 @@ def _get_ts_parameters(times, ts_data, ts_time, ts_z, z_sorted=False):
...
             y = y[ok_indices]
             z = ts_z[ok_indices]
+            y_error = y_error[ok_indices]
             # initial guess
             i = y.argmax()
             guess = [y[i], z[i], (z.max() - z.min()) / 3]
             # actual fit
             try:
-                _, _, psigma = gaussian_fit(z, y, guess)
+                _, pmean, psigma = gaussian_fit(z, y, guess, y_error, True)
```
gtrevisan, line 1051: let's use named args, here.

gtrevisan (changes requested):
cool! minor changes, let's keep stuff within the loop at a minimum.

gtrevisan (approved):

gtrevisan:
hmm, weird, one test now fails, please investigate -- rerunning did not help.
```
poetry run pytest -v tests/test_output_setting.py -k test_batch_csv
```
there is now a conflict of object vs integer:
```
FAILED tests/test_output_setting.py::test_batch_csv - AssertionError: Attributes of DataFrame.iloc[:, 2] (column name="commit_hash") are different
Attribute "dtype" are different
[left]:  object
[right]: int64
```

yumouwei:
Wait this is confusing...

- Previously `test_batch_csv` was failing on github but it passed on both MFE & omega
- Now it passes on github but fails on MFE and omega

`FAILED tests/test_output_setting.py::test_batch_csv - AssertionError: Attributes of DataFrame.iloc[:, 2] (column name="commit_hash") are different`

gtrevisan:
hm, it does pass on dev, though...

yumouwei:
I'm checking `retrieval_manager.py`. Looks like the `commit_hash` column is converted from `string` to `dtype: object` at a certain step.

gtrevisan:
there might be a missed bytes to string conversion somewhere, then.

yumouwei:
Nah it's dumber than that: if `commit_hash` is entirely made of numbers, then `pd.read_csv(csv)` will assume the column to be `dtype=int` instead of `string`. All I need to do is to add `pd.read_csv(..., dtype={'commit_hash': 'object'})`.

gtrevisan:
okay, following my thoughts on the other PR, I feel we should merge this one as it was, without your last commit.
we will discuss a better fix in the other thread.

that said, why did our tests originally fail with your commit https://github.com/MIT-PSFC/disruption-py/pull/402/commits/ad531524fdb34a84e787f11c8b4d27cd69093ac7, which did include two non-numeric characters?

yumouwei:
I was checking if these spiky signals also affect the Te peaking factor. It turns out that the Te PF computation only considers the channels above the midplane when it maps the Te measurements to the equally-spaced array, so the problematic channels (which are all located below the midplane) get ignored. Note that this is different from the Te PF method in the DIII-D branch, which maps the TS channels to the flux coordinate first before interpolating them onto a equally-spaced spatial basis. In that case, every channel would be considered in the calculation.

For now everything is fine, but we should keep this in mind when we modify this function in the future.

image

PR #405, opened 2025-02-24
Specify commit_hash dtype in all pd.read_csv calls

yumouwei:
- Added `dtype={"commit_hash": str}` argument in all occurrence of `pd.read_csv` except those in the draft folder.
- The csv file dumped by disruption-py should always have the `commit_hash` column regardless of the retrieval settings, therefore I didn't add a statement to check if the column exist.


gtrevisan:
we are probably going to go with option 3, that is, dropping it entirely.

Issue #269, opened 2024-08-20
Use xarray internally


amdecker:
## Xarray Notes
- [ ] Update documentation to refer to xarray instead of Pandas
- [ ] Add xarray formats for cache & output

There will need to be lots of small changes to the code all over the place if we want to use xarray instead of Pandas. This table gives an overview of pandas methods we use and their equivalent xarray method. 

Pandas method | Files used | xarray equivalent/notes
-- | -- | --
to_csv | output_setting.py, pytest_helper.py | N/A, but can export to pandas
read_csv | cmod physics.py, shotlist_setting.py, output_setting.py | N/A, but can export to pandas
read_sql_query | sql.py | N/A, but we can use pyodbc (which we already have wrappers for in DisruptionPy)
concat | all over | https://docs.xarray.dev/en/stable/generated/xarray.concat.html
pandas has interpolation, but we are using scipy's interpolate | lots of physics methods use our custom interpolation | https://docs.xarray.dev/en/stable/generated/xarray.DataArray.interp.html
merge_asof(on="time",                direction="nearest",...) | retrieval_manager.py and eval_against_sql.py | something like y.reindex(time = x.time, method="nearest", ...) reindex docs
df.columns | caching.py, and so many other places | ds.data_vars, unless we need one of the coordinates, like time, in which case we would use ds.coords
df["col"].values | all over | Same! ds["data_var"].values
np.isnan | runner.py and more | Do np.isnan(ds["data_var"].values) or ds["data_var"].isnull()
assignment of columneg method_dict[parameter] = np.full(len(pre_filled_shot_data), np.nan) | runner.py and more | Either ds[parameter].values or asssign with coordseg ds[parameter] = ('time', new_values)
len(df) | runner.py and more | Pandas will give the number of rows when you call len(), but xarray will give the number of data variables. So instead it would be best to get the length of a specific dataarray like len(ds.shot_id) or len(ds.time)
df.drop(column) | retrieval_manager.py and more | Same! ds.drop(column)
df.to_string() | data_difference.py | We could use str(ds), or if we want to keep things in a nice table format, just convert to Pandas
df.iloc | data_difference.py | ds.parameter.loc or if you want to select by coordinate index could do eg ds.isel(time=[0, 2, 4, 6])
pd.testing.assert_frame_equal | data_difference.py | xr.testing.assert_equal

gtrevisan:
closed with:
- #442

PR #407, opened 2025-02-27
Use xarray instead of pandas internally

amdecker:
EDIT by GLT:

- disruptionpy returns xarray dataset with dimensions shot and time
- keep old-school dataframe option available (for now)
- xarray datasets are saved to netcdf4/hdf5, while pandas dataframes save to csv
- `DataFrameOutputSetting` subclasses `DatasetOutputSetting`
- `SQLOutputSetting` subclasses `DataFrameOutputSetting`
- removed `ListOutputSettings`, `DictOutputSettings`, `OutputSettingsDict`
- removed `CSVOutputSetting`, `BatchedCSVOutputSetting`
- removed `CacheSettingDict`, `DFCacheSetting`, `CSVCacheSetting`
- removed some utilities
- converted some physics methods to return a native xarray dataset
- dropped commit_hash (for now)
- revamped dataframe assertions
- simplified column down-selection

to do:
- [x] merge and drop `hdf5` extras from pandas
- revamp output and cache settings to only support xarray datasets and sql
- understand and reabsorb difference between `cache_data` and `pre_filled_shot_data`
- investigate all places with lingering pandas logic, even in absence of calls to pandas
- fully revamp testing with sql
- fully revamp i/o with sql
- [x] switch back default to dataframe
- [x] include deprecation for dataframe support
- update documentation
- [x] final cleanup of a/an, xr/xarray, dataset/Dataset, ...

example from our `simple` workflow with the test shotlist for C-MOD.
before, using a poorly-indexed pandas dataframe:
```
       time        shot commit_hash        i_efc  ...   tau_rad  te_core_vs_avg_ece  te_edge_vs_avg_ece  te_width_ece
0    0.0600  1150805012     f978ae1 -1073.678169  ... -0.202146                 NaN                 NaN           NaN
1    0.0800  1150805012     f978ae1  -483.126275  ... -0.072099                 NaN                 NaN           NaN
2    0.1000  1150805012     f978ae1  -953.190392  ... -0.058194                 NaN                 NaN           NaN
3    0.1200  1150805012     f978ae1 -1572.869141  ...  0.008998                 NaN                 NaN           NaN
4    0.1400  1150805012     f978ae1 -1801.871277  ... -0.117796                 NaN                 NaN           NaN
..      ...         ...         ...          ...  ...       ...                 ...                 ...           ...
946  1.6405  1150805021     f978ae1 -1017.803607  ...  0.107822                 NaN                 NaN           NaN
947  1.6415  1150805021     f978ae1 -1411.955629  ...  0.006596                 NaN                 NaN           NaN
948  1.6425  1150805021     f978ae1 -1927.148726  ...  0.014654                 NaN                 NaN           NaN
949  1.6435  1150805021     f978ae1 -2477.766338  ... -0.087103                 NaN                 NaN           NaN
950  1.6445  1150805021     f978ae1 -2698.175370  ...  0.030866                 NaN                 NaN           NaN

[951 rows x 64 columns]
```
and after, using a self-described multi-dimensional xarray dataset:
```
<xarray.Dataset> Size: 1MB
Dimensions:               (shot: 10, time: 294)
Coordinates:
  * time                  (time) float64 2kB 0.06 0.08 0.1 0.1 ... 1.76 1.78 1.8
  * shot                  (shot) int64 80B 1150805012 1150805022 ... 1150805021
Data variables: (12/61)
    i_efc                 (shot, time) float64 24kB ...
    p_rad                 (shot, time) float64 24kB ...
    dprad_dt              (shot, time) float64 24kB ...
    p_lh                  (shot, time) float64 24kB ...
    p_icrf                (shot, time) float64 24kB ...
    p_input               (shot, time) float64 24kB ...
    ...                    ...
    te_core_vs_avg_ece    (shot, time) float64 24kB ...
    te_edge_vs_avg_ece    (shot, time) float64 24kB ...
    te_width_ece          (shot, time) float64 24kB ...
    n_e                   (shot, time) float64 24kB ...
    dn_dt                 (shot, time) float64 24kB ...
    greenwald_fraction    (shot, time) float64 24kB ...
```

references:
- https://pandas.pydata.org @ https://pandas.pydata.org/docs/user_guide/
- https://xarray.dev @ https://docs.xarray.dev/en/stable/user-guide/

closes
- #269


disruption_py/settings/output_setting.py
```diff
@@ -159,18 +166,6 @@ def _output_shot(self, params: OutputSettingParams):
         """
         return [s.output_shot(params) for s in self.output_setting_list]
 
-    def stream_output_cleanup(self, params: CompleteOutputSettingParams):
```
gtrevisan, line 162: do we need `OutputSettingList` and `OutputSettingDict`, at all?

tests/test_output_setting.py
```diff
@@ -53,10 +54,10 @@ def initial_mdsplus_data_fixture(shotlist, tokamak, test_file_path_f) -> Dict:
     """Get data in multiple formats"""
     output_settings = [
         "dataframe",
-        "list",
-        "dict",
+        "dataset",
         test_file_path_f(".csv"),
         test_file_path_f(".hdf5"),
+        test_file_path_f(".nc"),
```
gtrevisan, line 60: we might not need all of these cases, I'm not even sure there is any difference between xarray saving a file with `h5`/`hdf5`/`nc`/`cdf` extension...

tests/test_output_setting.py
```diff
@@ -85,21 +86,28 @@ def test_output_exists(initial_mdsplus_data, test_file_path_f):
...
     assert isinstance(sql_output, pd.DataFrame), "DataFrame from SQL does not exist"
+    assert os.path.exists(test_file_path_f(".nc")), ".nc output does not exist"
     assert os.path.exists(test_file_path_f(".csv")), ".csv output does not exist"
     assert os.path.exists(test_file_path_f(".hdf5")), ".hdf5 output does not exist"
 
     assert_frame_equal_unordered(df_output, csv_output)
-    assert_frame_equal_unordered(csv_output, hdf_output)
-    assert_frame_equal_unordered(hdf_output, sql_output)
+    assert_frame_equal_unordered(df_output, sql_output)
+    xr.testing.assert_equal(ds_output, hdf_output)
```
gtrevisan, line 109: it feels to me that this is trivial since it compares the same in-memory representations from identical runs.
we might want to compare the in-mem with the dataset obtained by re-reading the output file, though.

gtrevisan (changes requested):
many thanks! :pray:

I'll comment here as I notice things in the next few days
if you are the chosen one to take care of some of this, great, otherwise no worries. :)

disruption_py/core/physics_method/runner.py
```diff
@@ -25,10 +25,10 @@
 from disruption_py.machine.method_holders import get_method_holders
 from disruption_py.settings.retrieval_settings import RetrievalSettings
 
-REQUIRED_COLS = {"time", "shot", "commit_hash"}
+REQUIRED_COLS = {"time", "shot"}
```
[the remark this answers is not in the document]
amdecker: It is still used here to make sure we return shot and time even when someone requests data for only a specific column. 
https://github.com/MIT-PSFC/disruption-py/blob/6b20641ba62f3be8894addb90ce650eafa553259/disruption_py/core/physics_method/runner.py#L382-L393

disruption_py/settings/output_setting.py
amdecker: If we want to be able to output data to multiple formats with one call to get_shots_data, then yes. If no one is going to use that feature, then there's probably no need for it.

disruption_py/core/physics_method/runner.py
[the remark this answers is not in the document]
gtrevisan: but aren't shot and time dimension coordinates, now?
I thought it would have all thought in terms of xarrays and only a few spots here and there added the "columns" for pandas.

disruption_py/settings/output_setting.py
gtrevisan: personally, I'd kill it, the only usable version is outputting both an object in memory and dumping the same object to file, but as I mentioned that should be a single class anyways, with a special constructor (eg `saveas="/path"`).
our internal functions to convert pd-xr and viceversa should be public for the users, though.

disruption_py/core/physics_method/runner.py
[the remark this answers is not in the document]
amdecker: Yes you are totally right! Right now, each DataArray only has the `time` coordinate, and this line with `REQUIRED_COLS` keeps the `shot` coordinate floating along with the Dataset as a whole. The shot coord is attached to the individual DataArrays in the OutputSetting. To remove the var `REQUIRED_COLS` we would need to add the `shot` coordinate when we construct the xarray datasets.

disruption_py/core/physics_method/runner.py
```diff
@@ -156,7 +108,7 @@ def filter_methods_to_run(
     """
     methods = retrieval_settings.run_methods
     if retrieval_settings.run_columns is not None:
-        columns = REQUIRED_COLS.union(retrieval_settings.run_columns)
+        columns = retrieval_settings.run_columns
     else:
         columns = None
```
amdecker, line 113: This can now be simplified to 
```
columns = retrieval_settings.run_columns
```
if we want a separate variable for `columns`. Otherwise we can remove the `columns` variables and just use `retrieval_settings.run_columns`

disruption_py/core/physics_method/runner.py
```diff
@@ -378,21 +346,8 @@ def populate_shot(
...
-                set(retrieval_settings.run_columns).intersection(
-                    set(local_data.columns)
-                )
-            )
-        )
-        local_data = local_data[include_columns]
-    return local_data
+    # column down-selection
+    if retrieval_settings.only_requested_columns and retrieval_settings.run_columns:
+        shot_data = shot_data[retrieval_settings.run_columns]
```
amdecker, line 351: awesome! much prettier

disruption_py/settings/cache_setting.py
```diff
@@ -80,141 +76,81 @@ def _get_cache_data(self, params: CacheSettingParams) -> pd.DataFrame:
...
-    ".hdf5": HDF5CacheSetting,
-    ".csv": CSVCacheSetting,
+    ".cdf": DatasetCacheSetting,
+    ".h5": DatasetCacheSetting,
+    ".hdf5": DatasetCacheSetting,
+    ".nc": DatasetCacheSetting,
 }
 
 
+# pylint: disable-next=too-many-return-statements
```
amdecker, line 142: This pylint disable might not be needed anymore

gtrevisan:
@yumouwei would you be able to update the main workflow image to mention Datasets, rather than DataFrames, and maybe even move EAST into the existing supported machines? :pray:

yumouwei:
DisruptionPy Workflow

What should I put under Storage?

gtrevisan:
I'd put `NetCDF4/HDF5, Zarr` for sure, maybe even `CSV` -- the first two are native xarray backends, while the last one can be thought of as the result of a dataframe conversion with stripped metadata information. your choice. thanks!

gtrevisan:
after a full DB test with shot and time as coordinates, I learned that this does not scale (at all!) due to the necessity of the allocation of the underlying numpy arrays.

for example, the resulting dataset of all our disruption warning shots for the shared time dimension would occupy:
**10,000 shots x 10,000 times** x 60 vars x 8 bytes/var ~ **50 GB** in RAM :red_circle: 

it seems more appropriate then to revert to something like a row index, which occupies:
**1,000,000 shot/time rows** x 60 vars x 8 bytes/var ~ **0.5 GB** in RAM :green_circle: 

since this impacts everything done here, including slicing, manipulation, and testing, it's probably better to keep this branch as guidance but re-implement all xarray features from scratch with the new architecture.

with a sad heart, I will close this branch for now. :skull:

Issue #406, opened 2025-02-26
Smoothing v_loop and p_oh


AlexSaperstein:
`v_loop` can be a very noisy signal on C-Mod, which in turn makes the `p_oh` signal often very noisy as well. when the plasma has significant ohmically heating (`p_oh` >~ p_aux), this then leads to a very noisy `f_rad` signal which can interfere with disruption prediction analyses. An an example of this is shown below for shot 1120105021, where the original `v_loop`, `p_oh`, and `radiated_fraction` disruption-py signals are in blue. The bottom plot also shows the comparison between `p_oh`, `p_icrf`, and `p_lh` to show how `p_oh` is comparable to p_aux.

![Image](https://github.com/user-attachments/assets/8bc8f444-6c00-4563-8f13-175b00b6a75c)

To mitigate this issue, I have been smoothing `v_loop` and then recalculating `p_oh` and `f_rad` from that smoothed signal. This is what is shown in the orange signals. Something similar could maybe be implemented in disurption-py.

gtrevisan:
thanks for the suggestion!

why don't you share a few snippets of your smoothing technique?
how sensitive is your desired result, in your experience, to the smoothing parameters?

this seems of broader interest than just p_oh, so maybe it's time to tackle how to setup generalized physics methods arguments.
not only in general, that is, passing arguments to physics methods, but also in the specific case of requiring signals to be smoothed over time.
let's table this for discussion at the next meeting!

AlexSaperstein:
Great!

The smoothing technique I use is just a simple box-car average with a 100ms window. Although it was just put together as a quick fix at the time, so something more sophisticated could also be used as well. The only requirement to consider is that the smoothing technique preserve causality. The box-car was just an easy way of implementing that, but there may be other better ones.

yumouwei:
I'm glad this one got brought up again. It is one of the things I'm planning to fix after #344 and #367.

So regarding vloop, there is another node (`efit_aeqdsk:vloopt`, currently called `v_loop_efit` in disruption-py) that is sampled at efit timebase and is smoother than the default `mflux:v0` used by the `get_ohmic_power` method. From the [efit documentation](https://nstx.pppl.gov/nstx/Software/Applications/EFIT_Signals.htm) this is supposed to be the measured loop voltage but I'd guess there's some additional smoothing done to it. 

![Image](https://github.com/user-attachments/assets/38b3e153-cfbe-4691-b98f-1970ec52ed5a)

[...]

PR #416, opened 2025-03-14
Improve CMOD ohmic power computation

yumouwei:
# Previous discussion

- #406 

# Implemented changes

- Replace the boxcar smoothing function for calculating `dip_smoothed` with the newly added `causal_boxcar_smooth` function.
- Use the same `causal_boxcar_smooth` function to smooth the noisy `v_loop` signal (from `\analysis::top.mflux:v0`).
  - For both `dip_smoothed` and `v_loop` I use 6-point boxcar smoothing which introduces a delay of about 0.5 ms observed at the end of the shot (see figures in the issue). I believe this amount of delay is tolerable since we are going to rerun EFIT with 1 ms interval.
- Replace all negative values of the computed `p_oh` to 0.
  - This isn't the ideal approach. I have also considered removing these data points and interpolate them from the nearby points. However, the nearby points aren't necessarily reliable either especially prior to a disruption where we observe higher level of fluctuation in magnetic signals due to plasma dynamics or noise. To me this is unavoidable since we are inferring the ohmic power through measurements on the plasma instead of the central stack voltage -- perhaps @granetz can comment on this.

# Note

- <s>I have **not** removed the plotting functions for comparing the raw & filtered signals!</s> I have removed the plotting functions. The previous version with the plotting functions are now moved to a new branch called `wei/improve-get-ohmic-plotting`. @AlexSaperstein please take a look using a couple other shots and make sure the levels of noise and delay are acceptable.

# TODO: 

- [ ] 🚨rebase🚨 to `origin/dev` in order to modify the docstring.
- [ ] The previous (non-causal) `smooth` function is also used in the EAST branch. I'll update those methods in a separate PR.

### DIFF
--- a/disruption_py/core/utils/math.py
+++ b/disruption_py/core/utils/math.py
@@ -87,7 +87,7 @@ def interp1(x, y, new_x, kind="linear", bounds_error=False, fill_value=np.nan, a
     return set_interp(new_x)
 
 
-def smooth(arr: np.ndarray, window_size: int) -> np.ndarray:
+def matlab_smooth(arr: np.ndarray, window_size: int) -> np.ndarray:
     """
     Implements Matlab's smooth function https://www.mathworks.com/help/curvefit/smooth.html.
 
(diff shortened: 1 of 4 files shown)


granetz:
@yumouwei @gtrevisan @hwietfeldt @AlexSaperstein The voltage signals for the C-Mod CS (which includes OH1, OH2U, & OH2L power supplies), and the #of turns in each of those solenoids is in the C-Mod engineering subtree for every shot, so in principle the CS stack voltage can be obtained.  Of course, this ignores any effect of the vacuum vessel.  But you could start by comparing the CS stack voltage against the derived loop voltage.

hwietfeldt:
Hey, I wanted to add myself to the conversation since I've learned a few things about Vloop (including \efit_aeqdsk:vloopt, which is currently used in case reading \mflux:v0 results in an error):

- From Bob: \mflux:v0 is not a measured signal. It’s from the flux loops data projected into the core (some linear combination of flux loops data plus derivatives)
- \efit_aeqdsk:vloopt is \mflux:v0 with a 20 ms smoothing window applied (window determined from the IAVEV input parameter to EFIT. I'm not sure if the smoothing is causal.
- If we run EFIT at a 1 ms sampling rate, we could set IAVEV to a value convenient for using smoothed loop voltage measurements output by EFIT. IAVEV does not enter into the equilibrium reconstructions.
- Bob also suggested potentially using Zeff (\spectroscopy:z_ave) as a way to back out the resistivity and calculate Vloop during flattop.  I believe Zeff is measured from a band of radiation that's typically devoid of line radiation and therefore dominated by Brehm. This was Earl's diagnostic, so he knows the details. To calculate the resistivity from Zeff, we'd also need to eliminate the Te dependence. I looked at a few shots via the scopes, and Zeff seems pretty noisy, so I don't think this is a path worth taking.

hwietfeldt:
IAVEV is an input to efit located in .NAMELISTS.EFITIN.IAVEV in the efit tree. Its comment in the node for IAVEV is "ms for averaging loop voltage, 20." For both analysis and efit18, IAVEV had a value of 20.

gtrevisan (approved):

Issue #415, opened 2025-03-13
Workflow interrupted by critical MDSplus NoData exception during shot setup


andrew-maris:
When trying to get all plasmas from Cmod, I get the below error that kills the data retrieval.

> 18:07:22.205 [CRITICAL] #930721017 | Failed to set up shot! %TREE-E-NODATA, No data available. Tree: analysis, Node: \analysis::efit:results:a_eqdsk:time

I have attached the full terminal output and error message below as a .txt. I am running on mfews04.

[terminal_cmod_plasmas_error.txt](https://github.com/user-attachments/files/19237557/terminal_cmod_plasmas_error.txt)

Below is the script:

```
#!/usr/bin/env python3

"""
Example usage of `get_shots_data` with all the default arguments explicitly assigned.
"""

from disruption_py.settings import LogSettings, RetrievalSettings
from disruption_py.workflow import get_shots_data

retrieval_settings = RetrievalSettings(
    cache_setting=None,
    efit_nickname_setting="default",#"default",#"disruption",
    # method/column selection
    # default None: all methods/columns
    # run_methods=None, #List of physics methods to run (default is None). If None, and run_columns is None, all methods will be run.
    # run_columns=None, # See above
    only_requested_columns=False,
    custom_physics_methods=[],
    # timebase settings
    time_setting="disruption_warning",#"efit",
    domain_setting="full",
    use_cache_setting_timebase=False,
    # not yet implemented
    interpolation_method="linear",
)

shot_data = get_shots_data(
    # required argument
    shotlist_setting="plasmas",#[1150805012, 1150805013, 1150805014],
    # default None: detect from environment
    tokamak="cmod",#None,#
    # default None: standard SQL/MDSplus connection
    database_initializer=None,
    mds_connection_initializer=None,
    retrieval_settings=retrieval_settings,
    output_setting="dataframe",
    num_processes=1,
    log_settings=LogSettings(
        # default None: "output.log" in temporary session folder
        log_file_path=None,
        file_log_level="DEBUG",
        log_file_write_mode="w",
        log_to_console=True,
        # default None: VERBOSE, or higher based on number of shots
        console_log_level=None,
        use_custom_logging=False,
    ),
)

# Save the shot data to a CSV file
shot_data.to_csv("cmod_shot_data.csv", index=False)
```

It looks like the issue is it's trying to get efit18 data for a shot that isn't in the tree. Is there a way to skip over shots like these?

yumouwei:
[...]

PR #418, opened 2025-03-18
Tolerate critical MDSplus errors

gtrevisan:
when a shot fails to setup, it yields a critical error.
the workflow now clearly displays those errors, but tolerates them better so that the data retrieval process can continue.

closes:
- #415


gtrevisan:
can you comment on what happens on `dev`, vs here?
this PR should only make the workflow tolerate MDSplus errors in the setup phase.

yumouwei:
In that case yes it does skip through the CRITICAL error due to missing EFIT tree (as seen in 930721017 ).

yumouwei (approved):

PR #427, opened 2025-04-02
Simplify time settings

gtrevisan:
- remove `IpTimeSetting`
- remove `SignalTimeSetting`

### DIFF
--- a/disruption_py/settings/time_setting.py
+++ b/disruption_py/settings/time_setting.py
@@ -467,99 +467,6 @@ class DisruptionTimeSetting(TimeSetting):
         return duration, signal_max
 
 
-class IpTimeSetting(TimeSetting):
-    """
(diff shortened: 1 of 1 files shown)


yumouwei:
I was talking with Enrique about detecting ELMs, and we thought it'd be helpful to fetch and interpolate data onto the `p_rad` timebase, which would be feasible with this `SignalTimeSetting` option. I did notice it is incomplete and not being available to any of the machine right now, so I wonder if we should either keep it or reimplement it in the future.

gtrevisan:
I'm happy to keep it if it's going to get used...
it should already work, though! please try it out and LMK.

yumouwei:
I just tested it and it does seem to work. However, right now it doesn't know how to trim out the start and the end of the shot if timebase of the specified signal extends over the actual discharge. We should keep this in mind next time we modify it.

PR #429, opened 2025-04-03
Fix non-git edge cases

gtrevisan:
- make the log header more resilient to non-git runs
- avoid importing the `tests` module which is unavailable for non-git installs
- log critical error if no shots or no processors

### DIFF
--- a/disruption_py/settings/log_settings.py
+++ b/disruption_py/settings/log_settings.py
@@ -163,12 +163,13 @@ class LogSettings:
         self.reset_handlers(num_shots=None)
 
         # header
-        package = "disruption_py"
+        package, *_ = __name__.split(".")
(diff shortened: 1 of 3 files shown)


gtrevisan:
I tested that I could run the simple example with a non-git installation by creating a small PEP-723 script:
```python
#!/usr/bin/env python3
# /// script
# requires-python = ">=3.11,<4.0"
# dependencies = ["disruption-py"]
# [tool.uv.sources]
# disruption-py = { git = "https://github.com/MIT-PSFC/disruption-py/", branch = "fix-non-git" }
# ///
from disruption_py.workflow import cli
cli()
```
and then running it with on-the-fly-venv creation with `uv`:
```
uv run test.py
```

Issue #408, opened 2025-03-03
Improve shot domain identification


yumouwei:
Disruption-py identify the shot domain (ramp up, flattop, or ramp down) by checking 2 conditions:

- `abs(dipprog_dt) <= TRESHOLD_DIPPROG_DT`
- `abs(ipprog) > THRESHOLD_IP_PROG`

For CMOD, the values of these two thresholds are currently: [EDIT by GLT]
- `TRESHOLD_DIPPROG_DT` = 1k
- `THRESHOLD_IP_PROG` = 100k

Both Alex and I have found another set of threshold values which appear to be more robust than these default values; in my case I used: [EDIT by GLT]
- `THRESHOLD_DIPPROG_DT` = 50k
- `THRESHOLD_IP_PROG` = 100k

in the ramping milestone project ([link](https://github.com/MIT-PSFC/onw_dev/blob/develop/onw_dev/generate_cmod_data/classify_disruption_domain.py) to script). We should take a look at this again and determine a best set of values.

In addition, I think it would be more useful to move domain identification from a setting to a physics method and returns the information as a categorical feature (e.g. 0=ramp up, 1=flattop, 2=ramp down). With this change the requested time base would be determined by `time_setting` alone, instead of by both `time_setting` and `domain_setting` as it's done currently. From a user's perspective, I'd say it's more convenient to store all data inside a single dataset and slice it to get the shot domain I'm interested in, instead of having 3 separate files and switch among them for each analysis.

## Related issue:
- #266

gtrevisan:
I edited your comment for clarity, please double check the numbers -- you adopted the same IP threshold but you raised the DIP threshold by 50 times, right?

it would be helpful to see some traces here and there and compare the logic.

@crea-psfc and @granetz should comment on the physics when they can, while I will tackle the data aspect.
on the latter, I agree, preprocessing should be kept at a minimum and that might very well be an output step function or even an xarray attribute.
let's table this for discussion at our next meeting, too.

granetz:
@yumouwei @gtrevisan @crea-psfc 1 kA/s is the value I used for the C-Mod flattop constraint in the disruption warning database

crea-psfc:
@yumouwei I agree in turning IP phases identification into a categorical variable.
Did you check your thresholds work also on DIII-D time traces? Should set these in device configuration?

yumouwei:
[...]

Issue #266, opened 2024-08-13
Inconsistent Flattop Calculations


amdecker:
### Matlab
Both flattop calculations are the same in each machine's `get_flattop_indices.m` script

https://github.com/MIT-PSFC/disruption-py/blob/5a13906aba52b22fc4119e7f867443d7936f3d03/CMOD/matlab-core/get_flattop_indices.m#L9-L11

https://github.com/MIT-PSFC/disruption-py/blob/5a13906aba52b22fc4119e7f867443d7936f3d03/DIII-D/get_flattop_indices.m#L9-L11

But in the `define_indices.m` scripts for each machine they are defined differently from each other and from the `get_flattop_indices.m` script.

https://github.com/MIT-PSFC/disruption-py/blob/5a13906aba52b22fc4119e7f867443d7936f3d03/CMOD/matlab-utils/define_indices.m#L16

https://github.com/MIT-PSFC/disruption-py/blob/5a13906aba52b22fc4119e7f867443d7936f3d03/DIII-D/utils/define_indices.m#L32-L36

And the `plot_Ip_error.md` script has another calculation:
https://github.com/MIT-PSFC/disruption-py/blob/5a13906aba52b22fc4119e7f867443d7936f3d03/CMOD/matlab-utils/visualization_code/plot_Ip_error.m#L25-L27

And this script for EAST defines the rampup as starting at `0.2` rather than the start or based on the programmed Ip time deriv 
https://github.com/MIT-PSFC/disruption-py/blob/5a13906aba52b22fc4119e7f867443d7936f3d03/EAST/disruption_warning_database.m#L138-L142

### DisruptionPy
Our DisruptionPy calculations for the two machines are also different. 

C-Mod flattop calculation:
https://github.com/MIT-PSFC/disruption-py/blob/6a1c64f25975bb76c627b5a1199691c157ebde2e/disruption_py/settings/domain_setting.py#L111-L113

DIII-D flattop calculation:
https://github.com/MIT-PSFC/disruption-py/blob/6a1c64f25975bb76c627b5a1199691c157ebde2e/disruption_py/settings/domain_setting.py#L174-L178

We define the rampup for C-Mod as everything up to the flattop (though this flattop calculation is different from the previous flattop calculations):
https://github.com/MIT-PSFC/disruption-py/blob/6a1c64f25975bb76c627b5a1199691c157ebde2e/disruption_py/settings/domain_setting.py#L202-L204

[...]

PR #433, opened 2025-04-07
Change domain settings to physics methods

yumouwei:
# Implemented changes

- <s>[x] Added `get_shot_domain` method to CMOD, D3D, and EAST physics methods </s>
  - [x] Added `ip_prog` as an output of the D3D `get_ip_parameters` method. `ip_prog` won't be tested in `test_against_cache.py`.
  - [x] Tested `get_shot_domain` for EAST shot. I'm currently having trouble with Remote-SSH to the EAST server
  - [x] Implemented generic version of `get_shot_domain`. This method returns a single `shot_domain` feature ~(1 = ramp up, 0 = flattop, -1 = ramp down)~ **(1 = ramp up, 2 = flat top, 3 = ramp down)**
- [x] Remove `domain_setting` from `RetreivalSettings`. EDIT by GLT: @gtrevisan 
- [x] Update docstrings
- [x] Update documentation

# Note

For the new `get_shot_domain` physics methods, I made the following modifications from the current `FlattopDomainSetting` functions:

- For the CMOD method, I used the new `threshold_dipprog_dt` value discussed in #408. I did not modify the threshold values for the D3D & EAST methods, and those should be investigated prior to releasing the curated datasets for those 2 machines.
- The method now finds the longest consecutive subsequence in `indices_flattop` and assign it as the actual flattop phase. This is done to account for cases such as CMOD 1150916006 and D3D 116177 which have 2 ramp up periods. For these cases I think the first shorter "flattop" period should be considered as part of the ramp up since `ip` hasn't stabilized and the shot hasn't reached its target scenario.
- All time points prior to the start of the flattop phase are considered as part of the ramp up, and all time points after the end of the flattop phase are considered as part of the ramp down, regardless of the actual sign of `dipprog_dt`. This accounts for shots that have a programmed overshoot of plasma current (e.g. CMOD 1120918023)
- If the method cannot find the flattop phase, then all time points would be labeled as the ramp up.
- With this new definition, a shot must have at least 1 point in the flattop phase in order to have a ramp down phase (i.e. a shot cannot have only a ramp up followed by a ramp down)

# Sample figures (using old shot domain indexing scheme)

image
image
image
image

# Related issues

- #266 
- #408

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
     steps:
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Install poetry
         run: pipx install poetry
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,7 +32,7 @@ jobs:
     steps:
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Install poetry
         run: pipx install poetry
@@ -50,7 +50,7 @@ jobs:
         run: poetry run mkdocs build --site-dir _site
 
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-pages-artifact@v4
 
   deploy:
     if: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,7 +26,7 @@ jobs:
         tool: [black, isort, pylint, ruff, toml-sort, yamllint]
     name: ${{ matrix.tool }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Install poetry
         run: pipx install poetry
@@ -51,7 +51,7 @@ jobs:
       github.event.pull_request.draft == false
     steps:
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Install shellcheck
         run: sudo apt-get install shellcheck
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
     steps:
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Add MDSplus sources
         env:
--- a/poetry.lock
+++ b/poetry.lock
@@ -244,14 +244,14 @@ extras = ["regex"]
 
 [[package]]
 name = "beautifulsoup4"
-version = "4.13.4"
+version = "4.13.5"
 description = "Screen-scraping library"
 optional = false
 python-versions = ">=3.7.0"
 groups = ["lab"]
 files = [
-    {file = "beautifulsoup4-4.13.4-py3-none-any.whl", hash = "sha256:9bbbb14bfde9d79f38b8cd5f8c7c85f4b8f2523190ebed90e950a8dea4cb1c4b"},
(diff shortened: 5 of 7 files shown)


yumouwei:
I was checking the MATLAB branch, and it looks like the original CMOD `get_flattop_indices.m` used `threshold_dipprog_dt` = 60 kA/s, which is very close to the value I found which is 50 kA/s, instead of 1 kA/s as in the current `domain_setting.py`. I wonder why this was changed during the translation.

https://github.com/MIT-PSFC/disruption-py/blob/5d1a80dbefe7785f22db77d3c6c0c07a1d964701/CMOD/matlab-core/get_flattop_indices.m#L8-L11

gtrevisan:
as discussed in the group meetings, please switch up the current legend, which is insightful but possibly more confusing, into a more standard ordered list of integers, either 0, 1, 2 or 1, 2, 3.

then, let's tag someone for a more hands-on review which should include some actual runs on a handful of shots.

I will take a deeper look at the framework-wide changes, including a possible removal of the equivalent setting.

yumouwei:
shot_domain

1150724003 & 005 look like interesting cases. Right now the `get_shot_domain` method is designed to find the longest continuous sequence from candidate flat top time points and then assign that as the actual flat top phase. This was added from the previous implementation to deal with 2 stage ramp up such as cmod 1150916006 and d3d166177 (top of the page). I don't know why these 2 shots have programmed ip like this, and I couldn't find any info from the logbook either.

yumouwei:
shot_domain (1)

Example d3d shots.

Note that we **did not** modify the shot domain identification thresholds for D3D or EAST in this PR. The only thing changed from the previous `domain_setting` is to find the longest continuous sequence and assign that as the flattop.

- 165128 had a mild ramp in `ip_prog` over the entire shot, which tricks the method to consider the entire shot as ramp up.
- 176162 also had a mild ramp after the identified ramp up phase.
- 172449 & 173884 both had a 2 stage ramp up. I'm really curious why people decided to do this type of shot style. 

Also curiously, the thresholds in `domain_setting` are actually different from what they were in Bob's MATLAB scripts. I wonder who made those changes and what was the intention. The new `ip_prog` threshold I found for CMOD (#408) are actually closer to Bob's original values.

| threshold | MATLAB | domain_setting.py | this PR |
| --- | --- | --- | --- |
| cmod ip_prog | $\le$ 6e4 | $\le$ 1e3 | $\le$ 5e4 |
| cmod dipprog_dt | > 1e5 | > 1e5 | > 1e5 |
| d3d ip_prog | $\le$ 6e4 | $\le$ 2e3 | (unchanged) |
| d3d dipprog_dt | > 1e5 | > 1e5 | (unchanged) |
| east ip_prog | $\le$ 1e3 | $\le$ 1e3 | (unchanged) |

I'd propose to change `d3d ip_prog` threshold back to Bob's value if @granetz supports it.

granetz:
William, I think your questions pertain specifically to distinguishing flattop from rampup and rampdown, yes?  I'm pretty sure that my specification for C-Mod's flattop was abs(dipprog_dt) < 1e3 (i.e. 1 kA/s).  I don't think I had any spec on ip_prog.

yumouwei:
@granetz Yeah I messed up the signal names, but I found these from the MATLAB branch and the conditions on `dipprog_dt` are different from what's in `domain_setting.py`

CMOD
https://github.com/MIT-PSFC/disruption-py/blob/5d1a80dbefe7785f22db77d3c6c0c07a1d964701/CMOD/matlab-core/get_flattop_indices.m#L9-L11

D3D
https://github.com/MIT-PSFC/disruption-py/blob/5d1a80dbefe7785f22db77d3c6c0c07a1d964701/DIII-D/get_flattop_indices.m#L9-L11

EAST
https://github.com/MIT-PSFC/disruption-py/blob/5d1a80dbefe7785f22db77d3c6c0c07a1d964701/EAST/utils/get_flattop_indices.m#L8

PR #439, opened 2025-04-09
Improve time settings

yumouwei:
# Implemented changes

  - GLT: introduced `_postprocess` for deduplication and unit conversion, 
- `EfitTimeSetting`
  - [x] Add `d3d_times()`.
  - [x] Create a generic get_efit_times method.
- `IpTimeSetting`
  - [x] Add `cmod_times` and `east_times`
  - [x] Fix unit conversion in `d3d_times`
  - [x] Add `use_efit_time_range` (default to False) option to remove time points outside the range of the EFIT timebase.
- `SignalTimeSetting`
  - [x] Add d3d wrapper function to allow using e.g. `tree_name="ptdata", signal_path="ip"`
- Other changes
  - [x] Fix all `np.where` output syntax #443 
  - [x] Figure out how to get the dimension from MDSplus to do unit conversion without hardcoding it based on the device.
  - [x] Set up test cases in `test_retrieval_settings.py`.

# Figures
image


disruption_py/settings/time_setting.py
```diff
@@ -250,6 +251,29 @@ def cmod_times(self, params: TimeSettingParams):
...
+            Array of times in the timebase.
+        """
+        return (
+            params.mds_conn.get_data(
+                r"\efit_a_eqdsk:atime",
+                tree_name="_efit_tree",
+                astype="float64",
+            )
+            / 1.0e3
+        )  # [ms] -> [s]
```
gtrevisan, line 275: this is one of those cases where black does not help out that much.
maybe this is better?
```python
time = params...
return time / 1e3
```

disruption_py/settings/time_setting.py
```diff
@@ -443,11 +467,13 @@ def _get_end_of_shot(cls, signal, signal_time, threshold=1.0e5):
...
         if threshold < 0:
             raise Warning("Threshold is negative.")
-        base_indices = np.where(signal_time <= 0.0)
+        (base_indices,) = np.where(signal_time <= 0.0)
         baseline = np.mean(signal[base_indices]) if len(base_indices) > 0 else 0
         signal -= baseline
         # Check if there was a finite signal otherwise consider the shot a "no plasma" shot
-        finite_indices = np.where((signal_time >= 0.0) & (np.abs(signal) > threshold))
+        (finite_indices,) = np.where(
+            (signal_time >= 0.0) & (np.abs(signal) > threshold)
```
gtrevisan, line 475: is there any simple way to reuse our previous lookup of `base_indices`, eg `~base_indices`?

disruption_py/settings/time_setting.py
```diff
@@ -551,13 +584,21 @@ def _get_times(self, params: TimeSettingParams) -> np.ndarray:
...
         except mdsExceptions.MdsException:
             params.logger.error(
                 "Failed to set up timebase for signal {signal_path}",
                 signal_path=self.signal_path,
             )
             raise
+        if params.tokamak == Tokamak.D3D:
+            signal_time /= 1e3  # [ms] -> [s]
+        if self.use_efit_time_range:
+            efit_time = EfitTimeSetting().tokamak_overrides[params.tokamak](params)
```
gtrevisan, line 596: I believe you should be able to just use `EfitTimeSetting().get_times()`...

disruption_py/settings/time_setting.py
```diff
@@ -551,13 +584,21 @@ def _get_times(self, params: TimeSettingParams) -> np.ndarray:
...
             )
-            return signal_time
         except mdsExceptions.MdsException:
             params.logger.error(
                 "Failed to set up timebase for signal {signal_path}",
                 signal_path=self.signal_path,
             )
             raise
+        if params.tokamak == Tokamak.D3D:
+            signal_time /= 1e3  # [ms] -> [s]
```
gtrevisan, line 594: is this always true? or is this true eg for EFIT signals, but really depends on the signal?
we can hardly say there's standardization there...

disruption_py/settings/time_setting.py
gtrevisan: also, I'm not sure we really need this crisscrossing of time settings, between signal and efit time bases...

yumouwei:
To get the unit of the timebase of a signal, I use

`params.mds_conn.get(f"units_of(dims_of({self.signal_path})), self.tree_name).data()`

However, this does not work for any PTDATA signal; e.g.

`params.mds_conn.get("units_of(dims_of(ptdata("ip", 161228))).data()`

returns a whitespace character (`' '`).

There is a PTDATA expression (`ptunits("ip", 161228)`) to get the unit of a signal, but I can't figure out how to wrap it to get the unit of the dimension of that signal.

disruption_py/settings/time_setting.py
yumouwei: The goal of this is to trim out the time points in the ip & other signals' timebase where there were no plasma, which I don't think anyone would be interested in. For these lines I've also tried getting EFIT time directly but that would require me to basically duplicate what's already in `EfitTimeSetting`, which is how I ended up doing it this way.

gtrevisan (commented):
cool! we should be able to talk in person and iron out the last details.

disruption_py/settings/time_setting.py
```diff
@@ -212,9 +212,42 @@ def __init__(self):
...
+        )
+        # Convert unit to seconds
+        time_to_sec = {"s": 1, "ms": 1e-3, "us": 1e-6}
+        if efit_time_unit.lower() in time_to_sec:
+            efit_time *= time_to_sec[efit_time_unit]
+        else:
+            params.logger.warning(
+                """Failed to get the timebase unit of EFIT tree {efit_tree_name}; 
+                assume the unit is in seconds""",
+                efit_tree_name=params.mds_conn.get_tree_name_of_nickname("_efit_tree"),
```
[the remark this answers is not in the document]
gtrevisan: it was here, it's currently scheduled to be removed from main since I dropped `CacheTimeSetting` from dev:
https://github.com/MIT-PSFC/disruption-py/pull/414/files#diff-72ebfed367f592c27f1bf6031a4434194951bcad7296a0dcc40426fa87ba49fbL248
we might want to think whether we can keep that, otherwise we might as well drop the `max_shot_time` config value, as well.

disruption_py/settings/time_setting.py
```diff
@@ -551,13 +665,41 @@ def _get_times(self, params: TimeSettingParams) -> np.ndarray:
             (signal_time,) = params.mds_conn.get_dims(
                 self.signal_path, tree_name=self.tree_name, astype="float64"
             )
-            return signal_time
+            # Remove duplicated time points
+            signal_time = np.unique(signal_time)
```
yumouwei, line 669: D3D shot 161228 `SignalTimeSetting(signal_path=r"\top.ech.total:echpwrc", tree_name="rf")` has a duplicated `t=0` at the end of the timebase. Not sure if this could happen to other timebase setting. What would be a more generalizable way to do this?

disruption_py/settings/time_setting.py
```diff
@@ -509,8 +564,34 @@ def d3d_times(self, params: TimeSettingParams):
...
+        Returns
+        -------
+        np.ndarray
+            Array of times in the timebase.
+        """
+        (ip_time,) = params.mds_conn.get_dims(r"\pcrl01", tree_name="pcs_east")
+        # For shots before year 2014, the PCRL01 timebase needs to be shifted
+        # by 17.0 ms
+        if params.shot_id < 44432:
+            ip_time -= 0.0170
```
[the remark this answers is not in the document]
yumouwei: https://github.com/MIT-PSFC/disruption-py/blob/5d1a80dbefe7785f22db77d3c6c0c07a1d964701/EAST/get_Ip_parameters.m#L78-L83

There's a lot of exceptions in EAST (e.g. where to look for btor or ip_prog for different shot ranges)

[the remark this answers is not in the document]
gtrevisan: hmm okay, not anything that would deserved to be referenced from our python script, then. NVM.

disruption_py/settings/time_setting.py
```diff
@@ -264,27 +225,24 @@ def east_times(self, params: TimeSettingParams):
...
-            Array of times in the timebase.
-        """
-        raise ValueError("EFIT timebase setting not implemented")
+        efit_time_unit = params.mds_conn.get_data(
+            r"units_of(dim_of(\efit_aeqdsk:ali))", tree_name="_efit_tree", astype="str"
+        )
+        efit_time_unit = efit_time_unit.lower()
+        # Convert unit to seconds
+        time_to_sec = {"ms": 1e-3, "us": 1e-6}
+        if efit_time_unit in time_to_sec:
```
gtrevisan, line 237: I think the `.lower()` was appropriate, here. doesn't hurt.

disruption_py/settings/time_setting.py
```diff
@@ -551,13 +610,39 @@ def _get_times(self, params: TimeSettingParams) -> np.ndarray:
...
             raise
+        # Convert unit to seconds
+        time_to_sec = {"ms": 1e-3, "us": 1e-6}
+        # get_unit_of_dim returns a whitespace character (' ') for D3D PTDATA signal
+        signal_unit = params.mds_conn.get_data(
+            f"units_of(dim_of({self.signal_path}))",
+            tree_name=self.tree_name,
+            astype="str",
+        )
+        signal_unit = signal_unit.lower()
```
gtrevisan, line 627: you can merge this onto the previous line

yumouwei:
`test_time_setting.py` passed on EAST

gtrevisan:
to do: 
- [ ] refactor all time parameters into machine configs,
- [ ] refactor common lines into new common functions,
- ...

PR #440, opened 2025-04-09
Simplify syntax to get data from PTDATA 

yumouwei:
This PR allows a user to use

 `conn.get_data('ip', tree_name='ptdata')`

to replace 

<s> `conn.get_data(f'ptdata("ip", {params.shot_id}, tree_name='d3d')` </s>
`conn.get_data(f'ptdata("ip", {params.shot_id}, tree_name=None)`

For the moment, this is only needed for using `SignalTimeSetting` with PTDATA signals on DIII-D. Therefore I did not modify any expression in `d3d/physics.py` or anywhere else. It also only works with simple expressions (i.e. it won't work with additional arithmetic operations).

### DIFF
--- a/disruption_py/inout/mds.py
+++ b/disruption_py/inout/mds.py
@@ -240,7 +240,7 @@ class MDSConnection:
 
         dim_nums = dim_nums or [0]
 
-        if tree_name == "ptdata":
+        if tree_name.lower() == "ptdata":
             path, tree_name = self._resolve_d3d_ptdata_path(path)
 
         if tree_name is not None:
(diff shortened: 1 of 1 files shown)


yumouwei:
I just found out that `conn.get_data(f'ptdata("ip", {params.shot_id}, ...')` doesn't need specifying `tree_name`. We should remove all the `tree_name="d3d"` or anything else to save the time used in opening those trees.

yumouwei:
In terms of using PTDATA signal in SignalTimeSetting, for now this is addressed by the `_get_times_d3d_wrapper` added in #439. We decided not to modify `mds.py` after the discussion above.

PR #411, opened 2025-03-04
Improve EAST physics methods

yumouwei:
# Improved physics methods:
- get_mirnov_std
- (get_prad_peaking) -- see note below
- get_n_equal_1_data
- get_n1rms_n2rms
- get_efit_gaps

# Newly added physics methods:
- get_radiated_power

# Remaining XFAILs:

- EFIT signals: `beta_n, beta_p, kappa, li, q0, qstar, q95, wmhd`: #380 
- `greenwald_fraction`, `p_oh`, `kappa_area`: computations are dependent on EFIT signals.
- `rad_input_frac`, `rad_loss_frac`, `p_input`: computations are dependent on `p_oh` which uses EFIT signals.
- `beta_p_rt`, `li_rt`, `wmhd_rt`: fresh data are all NaN, but there are cached data in the `disruption_warning` table. I don't know where those data were obtained from.

# Notes on improved/added methods
## 1. get_mirnov_std
- Fix data address & std computation
- Result is still slightly different from SQL table (and MATLAB). This is because for each time point the method computes the standard deviation within a time window; because of the high sampling rate of the mirnov sensor and the floating point bit error, sometimes this window may include one extra or one less data point (e.g. 101 or 99 instead of 100). This does not significantly affect the outputs (difference is <0.5%) but it does trigger `test_against_cache`. For this reason, I kept both `mirnov_std` and `mirnov_std_normalized` in the `expected_failure_columns`.

| shot  | time | fresh (disruption-py) | fresh (matlab) | stored (sql table) |
|-------|------|-----------------------|----------------|--------------------|
| 55012 |  0.2 |              0.006171 |       0.006171 |           0.006171 |
| 55012 |  0.3 |              0.002339 |       0.002329 |           0.002339 |
| 55012 |  0.4 |              0.002952 |       0.002965 |           0.002965 |
| 55012 |  0.5 |              0.002942 |       0.002964 |           0.002964 |
| 55012 |  0.6 |              0.002022 |       0.002015 |           0.002022 |

Note that neither disruption-py, matlab, nor the sql table match another column completely. Technically I can change the time window to a fixed index-based window but the difference would be too minuscule to matter.

## 2. get_radiated_power (translated from `prad_bulk_xuv2014_2016.m`)
Both this method and `get_prad_peaking` use data from the AXUV array. However the ways these two methods deal with what's noted as "bad" channels (ch 12 and 16) are different.

- For `get_radiated_power`, the orders are: 
[...]

PR #451, opened 2025-04-30
Temporarily add expected failures for EAST

gtrevisan:
a couple of MDSplus trees for EAST suddenly became unavailable, so we temporarily add the derived features to the expected failure list.

impacted trees:
- `east`
- `east_1`

impacted features:
- `p_ecrh`
- `p_icrf`
- `p_lh`
- `p_nbi`
- `prad_peaking`
- `v_loop`

### DIFF
--- a/disruption_py/machine/east/config.toml
+++ b/disruption_py/machine/east/config.toml
@@ -37,7 +37,11 @@ expected_failure_columns = [
   "n_equal_1_mode",
   "n_equal_1_normalized",
   "n_equal_1_phase",
+  "p_ecrh", # east/east_1 tree issues
+  "p_icrf", # east/east_1 tree issues
   "p_input",
+  "p_lh", # east/east_1 tree issues
+  "p_nbi", # east/east_1 tree issues
   "p_oh",
   "p_rad",
   "parea",
@@ -46,6 +50,7 @@ expected_failure_columns = [
   "pkappa_area",
   "pq0",
   "pqstar",
+  "prad_peaking", # east/east_1 tree issues
   "q0",
   "q95",
   "qstar",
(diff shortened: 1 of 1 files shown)


yumouwei:
On the east branch there are 5 additional signals that are not working due to the missing `east` and `east_1` trees:

- n_equal_1_mode
- n_equal_1_normalized
- n_equal_1_phase
- rmp_n_equal_1
- rmp_n_equal_1_phase

These 5 signals are all returned by the `get_n_equal_1_data` method. They are marked as XFAILs on `dev`, and they are fixed in the to-be-merged `east` branch prior to the missing trees debacle.

Additionally, the following signals are not fetched from `east` or `east_1`, but are affected by the way the `get_power` method (in both `dev` and `east`) is implemented:

- `p_ecrh`
- `p_icrf`
- `p_nbi`

I think the better way is to fix the `get_power` method by adding try-except blocks, instead of adding them to XFAILs.

Furthermore `east` also needs to be rebased to `origin/dev` as it's out of date.

---

I'm trying to figure out what would be the correct order & places to implement these changes.

**Method 1:**
1. rebase `east` onto `dev` 
2. merge `tmp-fix-east` to `east` 
3. implement the above changes in `east`
4. merge `east` to `dev`

**Method 2:**
1. merge `tmp-fix-east` to `dev`
2. rebase `east` onto the new `dev`
3. implement the above changes in `east`
4. merge `east` to `dev`

(now it looks like they aren't that much different)

yumouwei:
Regarding the 3 power signals, all we need to do is to add a try-except block around every `get_data` line (e.g. L555). Right now `p_lh` fails and kills the entire method, which is why we are not getting `p_ecrh, p_icrf, & p_nbi` (from `analysis, icrf_east, nbi_east` respectively). I've checked these trees and they are still accessible.

https://github.com/MIT-PSFC/disruption-py/blob/b892d14e815ed2f95403f40bd889d5e2f4ebaaa6/disruption_py/machine/east/physics.py#L550-L561

gtrevisan:
oh, I see! makes sense then.

gtrevisan:
okay, @yumouwei, thanks for the clarifications! please add these changes into:
- #411

PR #460, opened 2025-06-02
Refactor C-MOD get_n_equal_1_amplitude

yumouwei:
# Original issue

The `get_n_equal_1_amplitude` method fetches data from 4 BP13 sensors and then does a least squares fit to compute the n=1 mode amplitude and phase. If there are less than 4 sensors available (e.g. shot 1010803001 which doesn't have the `BP13_GH` node), then the method terminates and returns all its outputs -- including `bt` which does not depend on the least squares fit -- as `np.nan`. 

# Implemented changes

The behavior of this method is now as follows:

- For the 4 BP13 sensors, if the magnitude of the mean value of a sensor's signal over the requested time domain is greater than 1 T, this sensor is then rejected as it was likely broken during the shot.
- To compute the n=1 mode:
  - If all 4 sensors are available, then the method will execute as before.
  - If there are 2 or 3 available sensors, then the method will log a warning (`"Only N of 4 BP13 sensors are available for calculating the n=1 mode"`) and continue to compute the n=1 mode.
  - If there are less than 2 available sensors, then the method will log a warning (`"Not enough sensors to calculate the n=1 mode (N of 4 available)."`) and return `n_equal_1_mode`, `n_equal_1_phase`, and `n_equal_1_normalized` as `np.nan`.
- The `bt` signal will always be returned regardless of the cases.


AlexSaperstein:
Are each of the sensors 90 degrees apart from each other? If not, it's important to check with N=2 and 3 to make sure at least two sensors are, otherwise I don't think an n=1 fit will necessarily be accurate

granetz:
C-Mod at 10-fold symmetry in the toroidal direction, i.e. horizontal and vertical ports at 36-degree intervals toroidally.  Therefore the toroidal spacing between BP sensor sets could only be multiples of 36 degrees, and so the 4 sets were spaced apart by 72, 108, 72, and 108 degrees.  Alex S is correct regarding the minimum number or working sensors required to get the amplitude of the n=1 component of Bp.

granetz:
William, I agree.

AlexSaperstein (approved):
looks good to me

AlexSaperstein:
> @AlexSaperstein do you think you will have some time for a physics-based review?

So I tried to start doing some vetting of this, but doing a good physics validation may be difficult considering the lack of n=1 MHD on C-Mod. The signals for the most part look generally fine, like those below for the sample shot referenced in the original issue (1010803001)

image

However, finding shots with n=1 activity in it to verify that the n=1 signals accurately show their growth/rotation is non-trivial. 

I got a short list of NTM shots from Zander, but the n-numbers corresponding to these NTM's have yet to be labeled, so it's unclear if we should expect to see n=1 signal from any of these. For now, maybe a more in-depth physics validation can be put on hold until that labeling has been done.

I will note, however, something I noticed taking a look at the signals for some of the shots in Zander's NTM database. Which is that when looking at some of the earlier shots (<2012), the n=1 signal and its normalized variant are unrealistically large (by like a factor of 1000), like for shot 1050415032 shown below.

image
 
I wonder if the units for some MDS nodes changed at some point?

yumouwei:
That 105x shot is interesting...My guess is that the unit of the probes were different in those days. I'll take a look later.

AlexSaperstein:
> I suggest looking at C-Mod shot 1160907005. It was a "locked mode calibration shot", with a locked mode starting at t=0.6 seconds and lasting for the remainder of the discharge. We sometimes ran locked mode calibration shots to calibrate the HIREX rotation velocity diagnostic. -- RSG

Great suggestion Bob!

I checked out the signals for this shot, and you can see a clear rise in the n=1 signal starting around 0.6 s. It's not as clean as I'd like, since the n=1 signal isn't flat beforehand, but this still seems good.

image

yumouwei:
image

For 1050415032 looks like there's a broken sensor which isn't marked as broken. I can set a threshold to check e.g. if the mean magnitude is greater 1 T over the shot and if so marks it as broken.

yumouwei:
image

For reference here are the BP13 signals from the shot Bob mentioned.

granetz:
Alex, William, thanks for making those nice plots. -- RSG

AlexSaperstein:
> I added a threshold check to reject shots such as 1050415032. @AlexSaperstein can you take a look at the other shots you found and see if they are either fixed or rejected?
> 
> I can also imagine a case where a sensor's signal stays in 0 during the entire shot. I didn't add this check but I'll do so if anyone has found an example shot of this.

Looks like they get rejected, and then the n=1 signals return nans since there are now < 3 working sensors.

AlexSaperstein (approved):

granetz:
Alex, William,
I looked in more detail at shot 1160907005 (the locked mode calibration shot that I pointed you to), and I think I might know why the n=1 signal was not zero prior to the locking at t=0.6 s.  As shown in the attached plot, there was a large, rotating 1/1 mode on each sawtooth.  Its frequency was ~4 kHz, so the rotation won't show up clearly in the magnetics signals since they were digitized at 5 kHz.  But it might show up in the n_equal_1_mode signal if that core mode coupled to the Bp sensors outside the plasma. -- RSG
(If the plot doesn't show here, let me know and I'll send it via email.)
n=1_on_sawteeth

PR #464, opened 2025-07-14
Update workflow figure

yumouwei:


gtrevisan (approved):
missing transparency? otherwise LGTM, thanks!

PS: we might have to wait a bit before merging due to the 1 kHz EFIT recomputation...

PR #466, opened 2025-08-12
Fix C-MOD te_width_ece unphysical values

hwietfeldt:
# Original Issue
The method `get_te_profile_params_ece` output unphysical values of `te_width_ece`. Output electron temperature profile widths were occasionally much larger than the plasma minor radius. This was discovered by @zapatace when assessing numerical differences in output data when casting data as np.float64 or np.float32. The unphysical values of `te_width_ece` were numerically unstable.  Often, the Gaussian profile fits failed, giving flat profiles with large, unstable standard deviations, from which `te_width_ece` is calculated.

# Implemented Changes
- Bug fix to initial guess of Gaussian fit: `(y.max() - y.min()) / 3` -> `(r.max() - r.min()) / 3`
  - The initial guess of the standard deviation should be based on variation in radial positions of the GPC channels, not temperature variation of the profile.
  - This fixed many of the bad profile fits that resulted in unphysical, numerically unstable outputs.
- Sanity check of output values:
  - Check that `te_width_ece` is positive and less than the minor radius.
  - If the above check fails, also set the peaking factors to NaN, since a bad fit indicates that the GPC2 diagnostic channels may not have been working properly for that time slice (misalignment, density cutoff, etc)
- Use only GPC2, instead of concatenating GPC1 and GPC2
  - @AlexSaperstein noted that he's seen discrepancies between the two GPC systems, which may have been leading to bad fits.
  - The GPC2 system had more spatial resolution (19 channels vs. 9) and was newer, so use GPC2.
- Ignore channels outside the SOL defined by magnetic EFIT
- Calculations performed on ECE time, then interpolated to requested time
  - With new requested timebase feature, now `get_te_profile_params_ece` can output meaningful scalars describing the Te profile at the ECE timebase (4 kHz), which may be useful for users interested in fast-timescale radiative collapses.
  - With 1 kHz EFITs, interpolating r0, and aminor onto the ECE timebase is safe, since they should be constant on such fast frequencies. 
  - This has nothing to do with bad fits, just another issue I noticed.

# Validation
I ran DisruptionPy for all C-Mod shots from 2012-2016. There were no critical errors, and the output values are in a reasonable range, as shown in the histogram below. Large Te profile widths (>0.2 m, close to the minor radius) are mostly output during ramp-up and ramp-down, and are generally rare. Note that around 45% of time slices had a NaN value for `te_width_ece` due to failing checks within the method. These checks are strict because the GPC2 system suffered from frequent artifacts and misalignments, particularly for shots with low Bt. 
te_width_histograms_new

I also looked at time traces of the signals output by `get_te_profile_params_ece` for a radiative collapse, shown below. The output signals (left) capture the edge cooling shown in the Te profiles I extracted at certain time slices (right). The edge has cooled by t=0.624 s before Te profile flattening immediately preceding the disruption (not sure if this is the flattening due to a global reconnection event or a coincidental sawtooth crash)
 
radiative_collapse

Similarly, the profile parameters tracked sawtooth crashes, as desired, for shots I manually inspected. Below is an example in which Te width and the core PF capture profile flattening during large sawtooth crashes. The edge PF doesn't change much during a sawtooth crash, which is expected.
large_sawteeth

There are some bad outputs that would need to be caught by the user. Below is an example of a density cutoff causing artificially low values of the Te width. Instances of density cutoffs are more common at lower Bt fields, which is why there is a threshold on Bt to output values (see related PR below). Because we don't have a density profile along the mid-plane, I'm not sure of a better way to get rid of time slices with density cutoffs but happy to take suggestions.
density_cutoff

There are also occasional instances in which bad channels mess up fits, as shown below.
bad_channels

In summary, the output te_width_ece is now in a physical range across shots and generally tracks changes to the Te profile. There are occasional errors, which is to be expected since ECE systems are tricky to use an inputs to automated routines due to many issues and artifacts, but the vast majority of time slices have a te_width_ece within a very reasonable range (0.8-0.15 m).

# Other Notes 
- I looked in MDSplus for an error in Te associated with each channel, like for Thomson scattering mentioned in (https://github.com/MIT-PSFC/disruption-py/pull/402)
- Comparing the Te profile from ECE to that from Thomson would be tricky since ECE measures a radial profile of the outboard mid-plane whereas Thomson measures a vertical profile at $R=R_0$. Additionally, the profile coverage differs. Therefore, I don't include a comparison in the automated routine.

# Related PR
- https://github.com/MIT-PSFC/disruption-py/pull/260


disruption_py/machine/cmod/physics.py
```diff
@@ -1559,9 +1547,8 @@ def _get_te_profile_params_ece(
           p. 053506, 2005, doi: 10.1063/1.1899311.
         - https://github.com/MIT-PSFC/disruption-py/pull/260
```
gtrevisan, line 1548: @hwietfeldt now that you submitted and got a number, you can add the current PR here as another list item.

disruption_py/machine/cmod/physics.py
```diff
@@ -1559,9 +1547,8 @@ def _get_te_profile_params_ece(
...
 
-        Last Major Update: Henry Wietfeldt (08/28/24), (PR: #260)
+        Last Major Update: Henry Wietfeldt (08/12/25), (PR: #)
```
gtrevisan, line 1550: @yumouwei what format/style did you use for further tweaks on methods? please suggest the same format to Henry.

disruption_py/machine/cmod/physics.py
```diff
@@ -1662,23 +1626,34 @@ def _get_te_profile_params_ece(
...
+
+                # Estimate Te width using Gaussian fit with center fixed on mag. axis.
+                # Fixed center is necessary bc we usually only see a half profile
+                guess = [y.max(), (r.max() - r.min()) / 3]
                 try:
                     pmu = r0[i]
                     _, psigma = gaussian_fit_with_fixed_mean(pmu, r, y, guess)
-                except RuntimeError as exc:
-                    if str(exc).startswith("Optimal parameters not found"):
+                except RuntimeError as e:
```
gtrevisan, line 1650: might as well leave it as `exc`, unless there was a reason...

disruption_py/machine/cmod/physics.py
```diff
@@ -1662,23 +1626,34 @@ def _get_te_profile_params_ece(
...
+                    if str(e).startswith("Optimal parameters not found"):
                         continue
-                    raise exc
-
-                # rescale from sigma to HWHM
+                    raise e
+                # rescale from sigma to HWHM for sigma values that are physical
                 # https://en.wikipedia.org/wiki/Full_width_at_half_maximum
-                te_hwhm[i] = np.abs(psigma) * np.sqrt(2 * np.log(2))
+                te_width[i] = psigma * np.sqrt(2 * np.log(2))
```
gtrevisan, line 1656: unless it's warranted, I'd keep the same name to maintain parallelism with our default TS Te width.
what is the impact of removing the absolute value?

disruption_py/machine/cmod/physics.py
```diff
@@ -1700,13 +1675,20 @@ def _get_te_profile_params_ece(
...
-        te_core_vs_avg = interp1(efit_time, te_core_vs_avg, times)
-        te_edge_vs_avg = interp1(efit_time, te_edge_vs_avg, times)
-        te_hwhm = interp1(efit_time, te_hwhm, times)
+        te_core_vs_avg = interp1(te_time, te_core_vs_avg, times)
+        te_edge_vs_avg = interp1(te_time, te_edge_vs_avg, times)
+        te_width = interp1(te_time, te_width, times)
+        # Sanity check: Te width should be positive and less than minor radius
+        # Unphysical Te width indicates ECE profile may be bad so set peaking factors to NaN too
+        aminor = interp1(te_time, aminor, times)
+        unphysical_mask = (te_width < 0) | (te_width > aminor)
```
gtrevisan, line 1684: if you remove the `abs`, wouldn't a negative width still be acceptable and only represent a symptom of the gaussian fit finding an unexpected local minimum?

gtrevisan (commented):
thanks, Henry! I left some comments, and I welcome William's review, as well...

yumouwei:
@hwietfeldt do you think if it's possible to define a validity range of an individual channel (e.g. < 5 keV) in order to reject the bad point shown in 1140725001? The [CMOD wiki](https://cmodwiki.psfc.mit.edu/index.php/Alcator_Parameters) puts the max core temperature at 8 keV so this can be tricky. Maybe you can figure out a way to check for outliers either in the profile at each individual time point, or in a channel over time.

hwietfeldt:
> @hwietfeldt do you think if it's possible to define a validity range of an individual channel (e.g. < 5 keV) in order to reject the bad point shown in 1140725001? The [CMOD wiki](https://cmodwiki.psfc.mit.edu/index.php/Alcator_Parameters) puts the max core temperature at 8 keV so this can be tricky. Maybe you can figure out a way to check for outliers either in the profile at each individual time point, or in a channel over time.

# Updates to Remove Outliers

I've added two checks to remove bad channels within a time slice as well as NaNing out time slices with bad profiles. They are summarized below.

## Check for artificially high channels

For the GPC2 system, there's a signal in MDSplus called te0, representing the central electron temperature. It seems to select a channel near the core with a valid temperature measurement (i.e. ignoring a channel near the core with an artificially high or low Te measurement, as in shot 1140725001 above). I couldn't find a script in which it automatically selected the channel, so instead I've added a check to remove channels with $T_e(r) > 2T_{e0}$. This fixes the problem in shot 1140725001, with the new result shown below (compare to the plot in my original post):
fix_outliers

## Check for density cutoffs

I've also added a rough check for density cutoffs to fix the problem with shot 1120808010 above. At sufficiently high densities, outer channels can suffer from density cutoffs, resulting in artificially low measured Te. The density cutoff occurs when the 2nd harmonic of the X-mode measured by GPC2 is less than the right-hand cutoff frequency of the plasma [1]: 

$$
\large{2\Omega_{ce}^2 < \omega_{pe}^2}
$$

After simplification, a given channel at radius $r$ is cutoff if 

$$
\large{n_e(r) > n_{crit}(r) = 2\epsilon_0 B^2(r) / m_e}
$$

We don't have a density profile, but we do have the line averaged density $\langle n\rangle$ through the core. For a parabolic profile with moderate to low peaking $\nu$ and $(r/a)^2 \ll 1$ such that

$$
\large{n(r) = n_0(1 - r^2/a^2)^\nu \approx n_0 (1 - \nu r^2/a^2)}
$$

Then 

$$
\large{n(r= a/\sqrt{3})\approx \langle n\rangle}
$$

For flat density profiles like in H-mode, this should also be fairly accurate. Instead of checking for a density cutoff at each channel, which we can't do since we don't have the density at each channel, I check to see if we would expect a cutoff at $r = a / \sqrt{3}$, and return NaN values for time slices in which we would expect a cutoff. For, a $B_0 = 5.4$ T pulse, the cutoff density is 4.2e20 $m^{-3}$, so the cutoff phenomena only occurs at very high densities on C-Mod.

This model makes many assumptions. In addition to the assumptions above, it also assumes that the line average density measured with the TCI, which has a vertical chord through the core, equals the line average density along the mid-plane (I think that's a fairly safe assumption). Despite these assumptions, it seems to work decently on shot 1120808010 as well as shot 116030034, which also suffers a density cutoff. Both plots are shown below.
 
density_cutoff_fix1
density_cutoff_fix2

## Updated validation

I re-ran `get_te_profile_params()` for all ~8000 C-Mod shots from 2012-2016. Below is a histogram similar to the one in my previous post. A similar fraction of time slices was returned as NaN (~47%), but the vast majority of time slices with dubiously low values of te_width_ece (< 0.06 m from 0.8 to 1.6 s on the right histogram in my original post) were removed, indicating that these two checks are generally removing profiles with artificially high or low channels without compromising good profiles.

te_width_histograms_outlier_cutoff_checks

[1] K. Zhurovich, et. al. "Calibration of Thomson scattering systems using electron cyclotron emission cutoff data," Rev. Sci. Instrum., vol. 76, no. 5, p. 053506, 2005, doi: 10.1063/1.1899311.

PR #470, opened 2025-09-02
Add option to choose the new "disruption" efit21 tree

gtrevisan:
add an option to switch to the new "disruption" EFIT tree:
- C-MOD: efit21 as a static tree,
- DIII-D: the efit21 equivalent (runtag `DISPY`).

for testing workflows, maintain the usual efit18 and efit18-equivalent trees, as tests would fail otherwise.

EDIT: efit21 is available for all shots, as opposed to efit18 which was recomputed only for disruptive shots.


hwietfeldt:
I visually compared several signals output by disruption-py using efit21 vs efit18 for 10 **C-Mod** shots with UFOs. The outputs seem to agree, though I noticed the outlier issue with p_oh and its derived signals described by @yumouwei above.

hwietfeldt (approved):

yumouwei (approved):

AlexSaperstein:
Everything looks good to me. I ran disruption-py for a typical disruption I've been looking at recently using both the default settings and by using efit18 (shown below).

image

It looks like disruption-py is correctly using efit21 by default now, and the signals look good.

I also tried pulling the radiative collapse shot-list, and had no issues (aside from it taking longer to do that now because of the larger number of samples).

AlexSaperstein (approved):
CLI worked well for me as well

Issue #394, opened 2025-02-02
Switch MDSplus to the mdsthin package


nathanielasun:
This library should be updated to work with the new "mdsthin" python thin client framework, which works with recent versions of python and numpy. This should be a relatively easy fix, as I was able to implement it by simply changing all imports to "mdsthin". Note: you will need to import exceptions directly from mdsthin and use as "exceptions".

gtrevisan:
hey there, thanks for reaching out!
it was already in the plan but it was missing from the roadmap, so I just added it there, too.

I did sit down with Stephen at the end of 2024 and tried `mdsthin` out, but it seemed to me that it was not ready for prime time just yet.
he did design it as a plug-n-play replacement, though, so I'm sure it's going to be working as intended really soon.

that said, the switch to mdsthin is not going to be implemented here in the immediate future.
one of the many reasons is that we support several machines, including those that still use an old version of MDSplus (eg EAST @ ASIPP).
you can have the latest client with the coolest APIs, but that client can support only a moderately-outdated server, so AFAIK something which uses a very old version won't be able to communicate properly with mdsthin.
furthermore, installing the required MDSplus packages on any linux box is straightforward, although it's not part of the documentation.
finally, in terms of numpy, last time I checked there was no widespread support for numpy2+ within the various servers (eg DIII-D @ GA).

feel free to reach out to me directly to discuss what you are doing, or planning to do!
we can support you independently of mdsthin.

PR #471, opened 2025-09-04
Add mdsthin as fallback from MDSplus

gtrevisan:
this PR introduces a two-fold approach to MDSplus APIs:

1. exploit the usual [full-fledged MDSplus installation](https://www.mdsplus.org/) (currently a prerequisite), or
2. fall back onto the [`mdsthin` package](https://github.com/MDSplus/mdsthin) (freshly included as a python dependency in the dev group).

the purpose is to be graceful and lenient in case system-wide MDSplus packages are missing (eg MAST), it's _not_ to support usage of `mdsthin`.

> [!CAUTION]
> do not run with `mdsthin`! it appears to be highly unstable for sizeable workflows.

close:
- #394

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2008,6 +2008,21 @@ files = [
 ]
 
 [[package]]
+name = "mdsthin"
+version = "1.6.1"
+description = "MDSplus Thin-Client implemented in pure python"
+optional = false
+python-versions = ">=3.7"
+groups = ["dev"]
+files = [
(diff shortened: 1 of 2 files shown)


samueljackson92 (approved):
Changes look good to me. I rebased my branch with MAST support off of this and these changes allow me to run disruptionpy without requiring `MDSplus` to be installed locally.

gtrevisan:
can you try with a full-scale test? every time I tried to run it via mdsthin it was freezing mysteriously after a few hundred shots...

PR #472, opened 2025-09-18
Cast all quantities to single precision by default

gtrevisan:
- remove `astype` argument from all MDSplus functions,
- remove `safe_cast` utility function,
- remove all cast operations from physics methods,
- have framework cast all dataframes to float32,
- have framework cast all times to float32,
- have framework cast down after reading test csv from disk.

NOTE: bad things might happen when we are reading units of MDSplus signals, if not in this current branch then whenever we will update to `numpy>2` as some changes were implemented in the underlying classes.

please help me out and test thoroughly!
- @yumouwei please review impact on physics methods changes,
- @zapatace please review impact on framework,
- @both/others, please test out with actual code runs.

if any developer wanted a specific physics method to return a double-precision float64 array, it can be done without framework changes by preparing a DataArray or Dataset and returning that one to the runner, which won't cast anything.


gtrevisan:
"casting to float32"? :confused: 

> For 161228, casting to float32 gives:
> Signal 	Value
> tmin 	0.10000000149011611938476562500000
> times[2192] 	0.10000000000000008881784197001252
> times[2193] 	0.10050000000000003375077994860476

zapatace:
I compared dataframes from all shots (10429 shots) in the disruption warning table using
(glt/float against dev) and (glt/float against main) branches. Most variables are identical. 🥳

For example,
image
image
image


A little history of the investigation and all plots:
[plots.zip](https://github.com/user-attachments/files/22558214/plots.zip)
[New glt_float branch PR.pdf](https://github.com/user-attachments/files/22558246/New.glt_float.branch.PR.pdf)

I'll approve the PR after this.

zapatace (approved):
After some testing this code reproduce an equivalent result as dev with C-Mod data. Therefore I approve it.

gtrevisan:
hmm... for me DIII-D tests show `h_alpha` XPASS for both dev and this branch.
instead, `dipprog_dt_rt` PASSes on dev but XFAILs here.

this would be a great time to take another look at all the XPASS and XFAILS... :eyes:

yumouwei:
You were probably using test-fast, which only uses shot 161228. 
For `h_alpha`:
- (dev) passes: 161228, 161237, 166177; fails: 166253
- (glt/float): all 4 shots passes.
You should see `h_alpha` fails in `dev` and passes in `glt/float` with the full test.

zapatace (approved):
All tests passed with same result again!

PR #473, opened 2025-09-22
Auto initialize dummy DB if SQL server is missing

gtrevisan:
automatically initialize an instance of the `DummyDatabase` object in case the SQL server (or database name) is missing.

this should be useful for both:
- MAST -- @samueljackson92 
- HBT-EP -- @yumouwei 

in the future, we should revamp both SQL and MDS initializers.


gtrevisan:
C-MOD test might look like:

```bash
# default, connects to DB
poetry run disruption-py
```

vs

```bash
# override DB, dummy object
DISPY_INOUT__SQL__HOST= poetry run disruption-py
```

PR #479, opened 2025-10-14
Fix DIII-D physics methods' errors

gtrevisan:
Fixes related to problems in EFIT01 data (e.g. `len(data) <= 1`)
- `get_efit_parameters`: check the length of a signal before computing its derivative
- `get_density_parameters`: improve logic for switching to `\d3d:denv2` for electron density data

Fixes related to other problems with D3D EFIT data
- All D3D physics methods: set `tree_name=None (default)` for all data retrieval operations from PTDATA to avoid missing tree error (e.g. `tree_name="_efit_tree"` may trigger "missing efit01 tree" error).
- `get_ip_parameters`: improve logic in determining invalid time points.
- `get_rt_ip_parameters`: improve logic in determining invalid time points.

Fixes on disruption-py logic and computation:
- `get_power_parameter`: replace `X*=1e3` operation with `X=X*1e3`
- `get_density_parameters`: fix typo in error checking
- `get_rt_ip_parameters`: add more details in logging
- `get_rt_ip_parameters`: fix logic in determining feedback_off_indices
- `get_peaking_factors`: fix logic in determining if TS data is available before computing the peaking factors
- `get_time_domain`: check if all utilized signals are available and valid before computing the time domain information.

test with:

```
disruption-py 172421 -m get_power_parameters

UFuncTypeError(<ufunc 'multiply'>, 'same_kind', dtype('float64'), dtype('int32'), 2)                                                                                                                                                                                   
```

```
disruption-py 170048 -m get_peaking_factors

TypeError("argument of type 'NoneType' is not iterable")                                                                                                                                                                                                                
```

```
[...]
```

I was not expecting to see a difference between:
```python
a *= b
```
and
```python
a = a * b
```
but there is -- eg: https://stackoverflow.com/q/38673531
thoughts? :eyes: 
do we need to change all other instances of in-place multiplication/division?

### DIFF
--- a/disruption_py/machine/d3d/physics.py
+++ b/disruption_py/machine/d3d/physics.py
@@ -823,7 +823,7 @@ class D3DPhysicsMethods:
             power_supply_railed[railed_indices] = 1
         except mdsExceptions.MdsException as e:
             params.logger.warning(
-                "power_supply_railed: Failed to get epsoff signal. Setting to NaN.",
+                "power_supply_railed: Failed to get epsoff signal. Setting to NaN."
             )
             params.logger.opt(exception=True).debug(e)
             power_supply_railed = np.full(len(params.times), np.nan)


disruption_py/inout/mds.py
```diff
@@ -312,8 +316,28 @@ def get_dims(
         logger.trace(shot_msg("Getting dims: {path}"), shot=self.shot_id, path=path)
         dims = [self.conn.get(f"dim_of({path},{d})").data() for d in dim_nums]
 
+        for i, dim in enumerate(dims):
+            dims[i] = self._check_mds_data(dim)
         return dims
 
+    def _check_mds_data(self, data) -> np.ndarray:
```
yumouwei: I also tried to add another logic to check `if data type is np.array and data is empty, return np.empty(0)` to fix the error found in `get_density_parameter`, but that causes more methods to fail.

disruption_py/machine/d3d/physics.py
```diff
@@ -407,15 +407,25 @@ def get_density_parameters(params: PhysicsMethodParams):
...
+            ne = [np.nan]
+            t_ne = [np.nan]
         # If EFIT disruption tree does not contain density data,
         # then read density from BCI subtree of D3D main tree
-        # TODO: Find a shot to test this logic
-        if len(~np.isnan(ne)) == 0:
+        # Example: 170480:
+        #  - r"\density" gives ne = array([], shape=(4, 0), dtype=float64),
+        #  - r"\denv2" gives actual density data
+        if sum(~np.isnan(ne)) == 0:
```
gtrevisan: I haven't tested it, but what about something like this?
```suggestion
        if not np.any(np.isfinite(ne)):
```

gtrevisan (commented):
minor comments, but otherwise LGTM. 👍 

are you able to provide a few execution lines to test this out on dev vs this branch? 🙏

gtrevisan:
there were also #170048 for `get_peaking_factors`, and `get_efit_parameters`, right? double check please?

gtrevisan:
:heavy_check_mark:

PR #500, opened 2025-12-18
Revive DIII-D get_n1_bradial_parameters

### LINKED ISSUE #261: DIII-D unaccessible PTDATA pointnames
# Description of the problem

The `get_n1_bradial_parameters` tries to access PTDATA DUSBRADIAL and (if it fails) ONSBRADIAL to fetch data for its computation. Currently both of these get data calls fail.
https://github.com/MIT-PSFC/disruption-py/blob/7284810809e9347b7244d3b4cab1b7dcaf0e7115/disruption_py/machine/d3d/physics.py#L705-L725

Corresponding lines in the MATLAB script
https://github.com/MIT-PSFC/disruption-py/blob/5a13906aba52b22fc4119e7f867443d7936f3d03/DIII-D/get_n1_bradial_d3d.m#L84-L99

I also tried checking these pointnames in ReviewPlus but it gave `ERROR %TREE-W-NNF, Node Not Found` error as well.
image

Also, on Omega I tried to check jTraverser by using tree name = "d3d" or "efit01" but neither of them worked. What's the proper tree name I should use?

yumouwei:
# Implemented changes

- Reimplement `D3DPhysicsMethods.get_n1_bradial_parameters`. This method was previously moved to `drafts/machine/d3d/physics.py` in v0.8 (#325) and later removed alongside all other draft scripts in v0.11 (#420).
- Create `D3DPhysicsMethods.get_btor` and use it in `get_n1_bradial_parameters` and `get_n1rms`.
- These new physics methods return `n_equal_1_mode`, `n_equal_1_normalized` (tested & XFAIL), and `btor` (not tested)

# Notes on testing
- `n_equal_1_mode` and `n_equal_1_normalized` are both marked as `XFAIL` as all 4 testing shots fall within the range of shots where the raw signal data need to be retrieved from files previously stored on `iris`. 
- I have checked these two signals with the last shot (177061) in the `DISRUPTION_WARNING` table and the outputs from this physics method match the data in the table.
- The outputs for shot 175552 (referenced in [Nucl. Fusion 59 (2019) 096016](https://iopscience.iop.org/article/10.1088/1741-4326/ab28bf)) do not match the stored data in the `DISRUPTION_WARNING` table. I found out that these stored data match those from the legacy DUD system (fallback data source in the method's logic) instead of the ONFR system which is available for this shot.

# Future TODO

- [x] Implement get data from storage for shots between 156199--172430 and 176030--176912 (described in #261).


disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
             z_cur_norm = z_cur / nominal_flattop_radius
         return {"zcur": z_cur, "zcur_normalized": z_cur_norm}
 
+    @staticmethod
+    @cache_method
+    @physics_method(columns=["btor"], tokamak=Tokamak.D3D)
+    def get_btor(params: PhysicsMethodParams):
```
gtrevisan, line 929: you might want to add attributes in the config for both methods.

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+
+        Returns
+        -------
+        dict
+            A dictionary containing the toroidal magnetic field (`btor`).
+
+        References
+        -------
+        - original sources: [get_n1_bradial_d3d.m](https://github.com/MIT-PSFC/disruption
+        -py/blob/matlab/DIII-D/get_n1_bradial_d3d.m), [get_n1rms_d3d.m](https://github.com
```
gtrevisan, line 946: have you verified that these line breaks are parsed normally by mkdocs?

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+                return {"n_equal_1_normalized": [np.nan], "n_equal_1_mode": [np.nan]}
+        t_n1 /= 1e3  # [ms] -> [s]
+        n_equal_1_mode *= 1.0e-4  # [G] -> [T]
+        n_equal_1_mode = interp1(t_n1, n_equal_1_mode, params.times)
+
+        # Calculate n_equal_1_normalized
+        b_tor = D3DPhysicsMethods.get_btor(params)["btor"]
+        if np.isnan(b_tor).all():
+            params.logger.warning(
+                "Failed to get b_tor signal to compute n_equal_1_normalized"
```
gtrevisan: I think the name of the physics method will be attached to error log by the framework, so this might be redundant...
put an intentional error somewhere to try it out!

gtrevisan:
@ZanderKeith, do you have some time to take a look?

disruption_py/machine/d3d/physics.py
yumouwei: We don't have attributes for d3d features yet. It'd be better to do them in a separate PR.

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+        -------
+        - original sources: [get_n1_bradial_d3d.m](https://github.com/MIT-PSFC/disruption
+        -py/blob/matlab/DIII-D/get_n1_bradial_d3d.m), [get_n1rms_d3d.m](https://github.com
+        /MIT-PSFC/disruption-py/blob/matlab/DIII-D/get_n1rms_d3d.m)
+        """
+        try:
+            b_tor, t_b_tor = params.data_conn.get_data_with_dims(
+                f"ptdata('bt', {params.shot_id})"
+            )  # [ms], [T]
+        except mdsExceptions.MdsException as e:
```
gtrevisan: I'm not sure we'd need this.

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+            # TODO: load data from NetCDF file
+            raise NotImplementedError
+
+        try:
+            # Get data from the ONFR system
+            n_equal_1_mode, t_n1 = params.data_conn.get_data_with_dims(
+                f"ptdata('onsbradial', {params.shot_id})",
+            )  # [G], [ms]
+        except mdsExceptions.MdsException:
+            try:
```
gtrevisan: this fallback would deserve at least a DEBUG statement, if not a warning!

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+                # Fallback: get data from the legacy DUD system
+                n_equal_1_mode, t_n1 = params.data_conn.get_data_with_dims(
+                    f"ptdata('dusbradial', {params.shot_id})",
+                )  # [G], [ms]
+                params.logger.verbose(
+                    "n_equal_1_mode: Failed to get ONSBRADIAL signal. Use DUSBRADIAL instead."
+                )
+            except mdsExceptions.MdsException:
+                params.logger.warning(
```
gtrevisan: this could be a CalculationError which would then downplayed as a warning by the framework.

disruption_py/machine/d3d/physics.py
```diff
@@ -931,19 +1039,13 @@ def get_n1rms_parameters(params: PhysicsMethodParams):
...
-            b_tor, t_b_tor = params.data_conn.get_data_with_dims(
-                f"ptdata('bt', {params.shot_id})"
-            )
-            t_b_tor /= 1e3  # [ms] -> [s]
-            b_tor = interp1(t_b_tor, b_tor, params.times)  # [T]
-            n1rms_norm = n1rms / np.abs(b_tor)
-        except mdsExceptions.MdsException as e:
+        b_tor = D3DPhysicsMethods.get_btor(params)["btor"]
+        if np.isnan(b_tor).all():
             params.logger.warning(
```
gtrevisan, line 1095: you might want to raise another CalculationError, here, and remember that shot number and physics method should be tagged on by the framework!

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
             z_cur_norm = z_cur / nominal_flattop_radius
         return {"zcur": z_cur, "zcur_normalized": z_cur_norm}
 
+    @staticmethod
+    @cache_method
```
gtrevisan: `@physics_method` already caches methods, so you can skip the explicit cache.

disruption_py/machine/d3d/physics.py
yumouwei: The fallback does raise a VERBOSE statement if it successfully gets data from DUSBRADIAL. The use of VERBOSE is consistent with the fallback in CMOD get_v_loop

https://github.com/MIT-PSFC/disruption-py/blob/4e35955d8eb2b0f70e31d0c51066e798dce32d24/disruption_py/machine/d3d/physics.py#L983-L990
https://github.com/MIT-PSFC/disruption-py/blob/4e35955d8eb2b0f70e31d0c51066e798dce32d24/disruption_py/machine/cmod/physics.py#L551-L557

gtrevisan: I missed that. but yeah, you should log to debug _before_ attempting the fallback, otherwise that might never get logged.

disruption_py/machine/d3d/physics.py
yumouwei: Do you mean removing this try-except block? If so then an error in this `get_btor` method would cause other methods that call `get_btor` (i.e. `get_n1_bradial_parameters` and `get_n1rms_parameters`) to raise an error and return all nans as well. I can avoid this by adding try-except blocks in those methods if that's a better way to do this.

disruption_py/machine/d3d/physics.py
yumouwei: It looks like only the shot number is automatically tagged to the error log if I intentionally poke an error. This applies to the comment on the other method as well

> 08:49:57.574 [WARNING] #201927 | Failed to get b_tor signal to compute n_equal_1_normalized                                                                                                
> 08:49:57.580 [WARNING] #201927 | Failed to get b_tor signal to compute n1rms_normalized

gtrevisan: because you are writing that yourself, but if you raise eg a CalculationError you will see the physics method name as well.

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +902,114 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+        try:
+            # Get data from the ONFR system
+            n_equal_1_mode, t_n1 = params.data_conn.get_data_with_dims(
+                f"ptdata('onsbradial', {params.shot_id})",
+            )  # [G], [ms]
+        except mdsExceptions.MdsException:
+            try:
+                # Fallback: get data from the legacy DUD system
+                n_equal_1_mode, t_n1 = params.data_conn.get_data_with_dims(
+                    f"ptdata('dusbradial', {params.shot_id})",
```
ZanderKeith: What is the difference between `dusbradial` and `onsbradial`? What do these signals mean / which coils are they coming from?

Towards the end of the shot they are usually in agreement but at the start they can be very different.

ZanderKeith (commented):
Methods look correct physics-wise. Ran them for a few shots from our recent DIII-D campaign with significant MHD activity. Signals line up well with those in reviewplus.

I'd still like to know what `dusbradial` vs `onsbradial` mean (and perhaps have a comment noting the difference), but other than that looks good to me.

disruption_py/machine/d3d/physics.py
yumouwei: I had this discussion with Cristina in 2024. To my understand the DUD system (which DUSBRADIAL is a part of) was a legacy machine warning system. It was replaced by the ONFR system around 2017/2018 and has not been actively maintained since then. I also noticed some shots have both DUSBRADIAL and ONSBRADIAL signals, and they don't agree with each other; I guess D3D people never bothered to retire the DUD system and it's still there eating up all their storage space.

- #261

gtrevisan:
I ran:

```
./scripts/disruption-errors.sh disruption_warning \
-m get_btor -m get_n1_bradial_parameters -m get_n1rms_parameters 
```

it "only" raised `NotImplementedError`... but for 10,797 shots out of 13,245 -- is this expected? at this scale?

yumouwei:
I just checked the d3d `disruption_warning` table for the number of shots within the `NotImplementedError` ranges, and it matches exactly with what you got.

```
SELECT COUNT(DISTINCT SHOT) AS shot_count
FROM disruption_warning
WHERE (SHOT BETWEEN 156199 AND 172430)
   OR (SHOT BETWEEN 176030 AND 176912)

SELECT COUNT(DISTINCT SHOT) AS total_shot_count
FROM disruption_warning

> shot_count = 10,797 , total_shot_count = 13,245
```

So yeah it's a lot of shots.

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +907,125 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+        - original source: [get_n1_bradial_d3d.m](https://github.com/MIT-PSFC/disruption-py
+        /blob/matlab/DIII-D/get_n1_bradial_d3d.m)
+        - pull requests: #[500](https://github.com/MIT-PSFC/disruption-py/pull/500)
+        - issues: #[261](https://github.com/MIT-PSFC/disruption-py/issues/261)
+        """
+        # The following shots are missing bradial calculations in MDSplus and
+        # must be loaded from a separate datafile
+        if 156199 <= params.shot_id <= 172430:
+            # Load data from custom tree structures on Omega
+            # TODO: check if running on Omega, otherwise raise an error
```
yumouwei: @gtrevisan is there a simple way to do this?

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +907,125 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+            params.logger.verbose(
+                "get_n1_bradial_parameters: Retrieving recomputed DUSBRADIAL signal "
+                "from custom MDSplus trees."
+            )
+            os.environ["bradial_path"] = config("d3d").physics.bradial_path.mds_trees
+            try:
+                tree = MDSplus.Tree("bradial", params.shot_id)
+                n_equal_1_mode = tree.getNode("dusbradial").data().squeeze()  # [G]
+                t_n1 = tree.getNode("dusbradial").dim_of().data()  # [ms]
+            finally:
```
yumouwei: @gtrevisan is there a better way to close the tree & unset bradial_path?

disruption_py/machine/d3d/physics.py
gtrevisan: I'd rather check whether the path you want exists.
there are other clusters at GA and in principle nothing is stopping users from running someplace else.

disruption_py/machine/d3d/physics.py
```diff
@@ -962,26 +967,46 @@ def get_n1_bradial_parameters(params: PhysicsMethodParams):
...
             params.logger.verbose(
-                "get_n1_bradial_parameters: Failed to get ONSBRADIAL signal. "
-                "Retrieving from DUSBRADIAL instead."
+                "get_n1_bradial_parameters: Retrieving recomputed DUSBRADIAL signal "
+                "from custom MDSplus trees."
             )
-            n_equal_1_mode, t_n1 = params.data_conn.get_data_with_dims(
-                f"ptdata('dusbradial', {params.shot_id})",
-            )  # [G], [ms]
+            os.environ["bradial_path"] = config("d3d").physics.bradial_path.mds_trees
```
gtrevisan: you can probably do this:
```suggestion
            os.environ["bradial_path"] = config(params.tokamak).physics.bradial_path.mds_trees
```
and maybe reapplying black.

disruption_py/machine/d3d/physics.py
gtrevisan: LGTM.
ideally we could read whether there already was a `bradial_path` in the user's env, and then place it back as it was after our modification, but it's probably overkill.

disruption_py/machine/d3d/physics.py
```diff
@@ -962,26 +967,46 @@ def get_n1_bradial_parameters(params: PhysicsMethodParams):
...
+            try:
+                tree = MDSplus.Tree("bradial", params.shot_id)
+                n_equal_1_mode = tree.getNode("dusbradial").data().squeeze()  # [G]
+                t_n1 = tree.getNode("dusbradial").dim_of().data()  # [ms]
+            finally:
+                tree.close()
+                del os.environ["bradial_path"]
+        elif 176030 <= params.shot_id <= 176912:
+            # Load data from a NetCDF file on Omega
+            # TODO: check if running on Omega, otherwise raise an error
```
gtrevisan: here, too, I'd just check existence of the path pulled from the config.

disruption_py/machine/d3d/physics.py
```diff
@@ -962,26 +967,46 @@ def get_n1_bradial_parameters(params: PhysicsMethodParams):
...
+            params.logger.verbose(
+                "get_n1_bradial_parameters: Retrieving recomputed DUSBRADIAL signal from recalc.nc."
+            )
+            filename = config("d3d").physics.bradial_path.recalc_nc
```
gtrevisan:
```suggestion
            filename = config(params.tokamak).physics.bradial_path.recalc_nc
```

disruption_py/machine/d3d/physics.py
```diff
@@ -962,26 +967,46 @@ def get_n1_bradial_parameters(params: PhysicsMethodParams):
...
+            ds = xr.open_dataset(filename)
+            shot_data = ds.sel(shot=params.shot_id)
+            n_equal_1_mode = shot_data["dusbradial_calculated"].values.copy()  # [G]
```
gtrevisan: I'm not sure you'll need to `.copy()` anything. do you?

disruption_py/machine/d3d/physics.py
```diff
@@ -962,26 +967,46 @@ def get_n1_bradial_parameters(params: PhysicsMethodParams):
...
-            n_equal_1_mode, t_n1 = params.data_conn.get_data_with_dims(
-                f"ptdata('dusbradial', {params.shot_id})",
-            )  # [G], [ms]
+            os.environ["bradial_path"] = config("d3d").physics.bradial_path.mds_trees
+            try:
+                tree = MDSplus.Tree("bradial", params.shot_id)
+                n_equal_1_mode = tree.getNode("dusbradial").data().squeeze()  # [G]
+                t_n1 = tree.getNode("dusbradial").dim_of().data()  # [ms]
+            finally:
+                tree.close()
```
gtrevisan, line 1005: I'm not sure you want the tree closing operation in the `finally` block... I'd just put it as a last thing in the `try` block, otherwise you will raise exceptions if eg the tree does not exist.

disruption_py/machine/d3d/physics.py
```diff
@@ -962,26 +967,46 @@ def get_n1_bradial_parameters(params: PhysicsMethodParams):
...
+                del os.environ["bradial_path"]
+        elif 176030 <= params.shot_id <= 176912:
+            # Load data from a NetCDF file on Omega
+            # TODO: check if running on Omega, otherwise raise an error
+            params.logger.verbose(
+                "get_n1_bradial_parameters: Retrieving recomputed DUSBRADIAL signal from recalc.nc."
+            )
+            filename = config("d3d").physics.bradial_path.recalc_nc
+            ds = xr.open_dataset(filename)
+            shot_data = ds.sel(shot=params.shot_id)
```
gtrevisan, line 1021: does that exist for all shots in the range, or should we check for existence and possibly catch KeyErrors and reraise them as our upcoming custom exceptions?

disruption_py/machine/d3d/physics.py
```diff
@@ -962,26 +967,46 @@ def get_n1_bradial_parameters(params: PhysicsMethodParams):
...
-                "Retrieving from DUSBRADIAL instead."
+                "get_n1_bradial_parameters: Retrieving recomputed DUSBRADIAL signal "
+                "from custom MDSplus trees."
             )
-            n_equal_1_mode, t_n1 = params.data_conn.get_data_with_dims(
-                f"ptdata('dusbradial', {params.shot_id})",
-            )  # [G], [ms]
+            os.environ["bradial_path"] = config("d3d").physics.bradial_path.mds_trees
+            try:
+                tree = MDSplus.Tree("bradial", params.shot_id)
```
gtrevisan: I'd put a statement here to log opening:
```suggestion
                tree = MDSplus.Tree("bradial", params.shot_id)
                params.logger.debug("Opened custom tree: {tree}", tree=tree.getFileName())
```

disruption_py/machine/d3d/physics.py
gtrevisan: AFAICS these 4 shots (between 176030 and 176912) are not included in the `.nc` file:
- 176108 176209 176226 176227

disruption_py/machine/d3d/physics.py
gtrevisan: AFAICS we have custom recalc trees for 3,814 shots out of the total 16,232 in the specified range -- 12,418 are missing.

yumouwei: This is consistent with the docstring in the MATLAB script

https://github.com/MIT-PSFC/disruption-py/blob/5d1a80dbefe7785f22db77d3c6c0c07a1d964701/DIII-D/get_n1_bradial_d3d.m#L27-L37

yumouwei: For that `finally` block, I was worrying if e.g. the `tree.getNode` statement raises an error and terminates this method, then this opened tree will never be closed. I don't know if the garbage collector does this automatically.

disruption_py/machine/d3d/physics.py
yumouwei: This is interesting. I didn't know xarray automatically protects the coordinates but not the data variables.

disruption_py/machine/d3d/physics.py
yumouwei: We should raise an `FetchDataError` here to make it consistent with the behavior of `xr.py`.

disruption_py/machine/d3d/physics.py
```diff
@@ -902,6 +906,136 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+                "get_n1_bradial_parameters: Retrieving recomputed DUSBRADIAL signal "
+                "from custom MDSplus trees."
+            )
+            # Check if the custom bradial tree exists
+            bradial_path = config(params.tokamak).physics.bradial_path.mds_trees
+            tree_path = os.path.join(bradial_path, f"bradial_{params.shot_id}.tree")
+            if not os.path.exists(tree_path):
+                raise FileNotFoundError(
+                    f"custom MDSplus tree does not exist: {tree_path}"
+                )
```
yumouwei: This and the one in L1000 could be a `FetchDataError` with the new custom error schema

gtrevisan:
- #555
- #554

disruption_py/machine/d3d/physics.py
gtrevisan: @yumouwei do those four shots have MDSplus custom trees, for example?

yumouwei: I checked the mds trees & the .nc file again:

- `recalc_bradial/bradial` folder contains 3814 shots between 156199--172430
- `recalc.nc` file contains 880 shots between 176030--176912

These 2 shot ranges do not overlap, so we don't need to do the extra check suggested above. 

One thing we can do is: for a shot that falls within these 2 shot ranges, first try to access the custom tree/.nc file; if that shot doesn't exist, then we fall back to checking the default mds trees; if that doesn't work again then we raise an error.

disruption_py/machine/d3d/physics.py
gtrevisan: absolutely!

disruption_py/machine/d3d/physics.py
gtrevisan: yes! great plan.

yumouwei: Hmm this turned out to be surprisingly hard because of all the `raise` statements we added and reliance on the framework to catch `get_data` errors. One thing I tried is to put the section on accessing `recalc_bradial` into a hidden function, so something like:

```
@staticmethod
def _retrieve_from_recalc_bradial(params: PhysicsMethodParams):
    ...
    return (n_equal_1_mode, t_n1)

def get_n1_bradial_parameters(...):
    if 156199 <= params.shot_id <= 172430:
        try:
            n_equal_1_mode, t_n1 = D3DPhysicsMethods._retrieve_from_recalc_bradial(params)
        except:
            n_equal_1_mode = t_n1 = None
    if 176030 <= params.shot_id <= 176912:
        # do something similar
    if n_equal_1_mode is None:
        # fetch from ptdata('onsbradial' or 'dusbradial')
```

but the `raise FetchDataError` statements in the hidden function aren't getting logged. I wonder if I'm overcomplicating things...

gtrevisan:
wait, wouldn't this work?

```python
            os.environ["bradial_path"] = bradial_path
            try:
                tree = MDSplus.Tree("bradial", params.shot_id)
            finally:
                os.environ.pop("bradial_path", None)
            params.logger.debug("Opened custom tree: {tree}", tree=tree.getFileName())
            try:
                n_equal_1_mode = tree.getNode("dusbradial").data().squeeze()  # [G]
                t_n1 = tree.getNode("dusbradial").dim_of().data()  # [ms]
            finally:
                tree.close()
                os.environ.pop("bradial_path", None)
```

yumouwei:
> wait, wouldn't this work?
> 
> ```python
>             os.environ["bradial_path"] = bradial_path
>             try:
>                 tree = MDSplus.Tree("bradial", params.shot_id)
>             finally:
>                 os.environ.pop("bradial_path", None)
>             params.logger.debug("Opened custom tree: {tree}", tree=tree.getFileName())
>             try:
>                 n_equal_1_mode = tree.getNode("dusbradial").data().squeeze()  # [G]
>                 t_n1 = tree.getNode("dusbradial").dim_of().data()  # [ms]
>             finally:
>                 tree.close()
>                 os.environ.pop("bradial_path", None)
> ```

I think it would. The reason why I ended up with that ugly nested `try-except` blocks is to raise different errors for `MDSplus.Tree` and `tree.getNode` and avoid duplicating the `os.environ.pop` statement.

Also, should we just remove those 2 `raise FetchDataError` and show only the `mdsException`?

gtrevisan:
I think we can leave it as per my suggestion, as it's short and clear and duplicates only one line.
native exceptions will be raised -- we can correct this in the future... if needed! but it'll probably be fine.

disruption_py/machine/d3d/physics.py
```diff
@@ -920,6 +924,142 @@ def get_z_parameters(params: PhysicsMethodParams):
...
+            bradial_path = config(params.tokamak).physics.bradial_path.mds_trees
+            tree_path = os.path.join(bradial_path, f"bradial_{params.shot_id}.tree")
+            if not os.path.exists(tree_path):
+                raise FetchDataError(f"Missing custom tree: {tree_path}")
+            os.environ["bradial_path"] = bradial_path
+            # Read from the custom bradial tree
+            try:
+                tree = MDSplus.Tree("bradial", params.shot_id)
+            finally:
+                os.environ.pop("bradial_path", None)
```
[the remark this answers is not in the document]
gtrevisan: good to know! done.

gtrevisan (approved):
@yumouwei please double check that my commits work as intended, and then we can merge! thanks.

yumouwei:
Go ahead & merge this. The `path=bradial_path` change looks like a way more elegant solution.

PR #501, opened 2026-01-05
Revamp installation readmes

gtrevisan:
revamp the installation instructions for clarity and to prepare for the upcoming release of the MAST open-access non-MDSplus workflow.

the installation section should now make it clear that there are four possible options:
- A) Run pre-installed
- B) Install as an application
- C) Install as a dependency
- D) Install from source

snippets are then provided for all options.

please review from the branch/file view, rather than the PR view:
- https://github.com/MIT-PSFC/disruption-py/blob/glt/docs/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ We recommend using either [Poetry](https://python-poetry.org/) or [uv](https://d
 ### Branches
 
 We suggest to install DisruptionPy:
+
 - from the stable branch, `main`, for production workflows;
 - from the development branch, `dev`, for development workflows;
 - from any other branch, only for testing workflows.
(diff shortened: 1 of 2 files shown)


yumouwei:
```
# if you use uv
uvx disruption-py

# if you use pipx
pipx install disruption-py
```
It looks like these two commands are technically not identical. According to Gemini, `uvx disruption-py` downloads the package, caches it in a virtual env, then runs the command immediately; `pipx install disruption-py` creates a permanent virtual environment and then installs the package in that environment. The equivalent commands are:

- `pipx run disruption-py` and `uvx disruption-py`
- `pipx install disruption-py` and `uv tool install disruption-py`

I tested `uvx disruption-py <shotno>` and `uv tool install disruption-py` on MFE and they worked as expected.

For pipx I did `pipx install disruption-py` and `disruption-workflow <shotno>` and got the following error:

> $ pipx install disruption-py
>   installed package disruption-py 0.9.0, installed using Python 3.10.12
>   These apps are now globally available
>     - disruption-workflow
> done! ✨ 🌟 ✨
>
> $ disruption-workflow 102709
> Traceback (most recent call last):
>   File "/home/yumouwei/.local/bin/disruption-workflow", line 3, in <module>
>     from examples.simple import cli
> ModuleNotFoundError: No module named 'examples'

yumouwei (approved):

PR #522, opened 2026-02-12
Dump configuration next to the output

gtrevisan:
this should help for reproducibility:

- dump configuration next to the outputs as JSON,
- drop `tests` structure (if not testing),
- drop `attributes` structure (as `nan` and `inf` are supported by TOML but not by JSON, and they should be stored as metadata anyways),
- add a debug statement.

in the future we might want to add this back in as xarray metadata, rather than a configuration dump.

### DIFF
--- a/disruption_py/workflow.py
+++ b/disruption_py/workflow.py
@@ -129,7 +129,7 @@ def get_shots_data(
         config_dict.pop("TESTS", None)
     with open(json_file_path, "w", encoding="utf8") as f:
         json.dump(config_dict, f, indent=3, sort_keys=True)
-    logger.debug("Dumped configuration into: {path}", path=json_file_path)
+    logger.verbose("Dumped configuration: {path}", path=json_file_path)
 
     database = _get_database_instance(tokamak, database_initializer)
     # Clean-up parameters


yumouwei (changes requested):
I'm able to generate the `config.json` file with `output_setting='dataset'` or `'dataframe'`. However, currently it does not work if I specify the file path (e.g. `output_setting="<path_to_file>/temp.h5"` or other format); it will only generate the dataset file but not the config file in the specified folder. 

Additionally, it would also be idea to set the name of the config file to something like `<dataset_file_name>.json` or `<dataset_file_name>_config.json` so that a user can distinguish them in case there are multiple dataset files stored in the same folder.

zapatace:
I tested different things...

First, I used to save files using output_setting and path. And I found that the json and log file are always saved in /tmp folder. Is that the expected behavior? 
For example:
```

shot_data = get_shots_data(
    tokamak="cmod",
    shotlist_setting=[1050420019],
    output_setting="./test.nc",
)
```
saves a file "test.nc” on the path expecified in output_setting but the json and log files are saved in the user's /tmp/. That's the problem @yumouwei had here https://github.com/MIT-PSFC/disruption-py/pull/522#pullrequestreview-3802167920.

Second, I also notice that if you rerun last code, you have the expected error
```
output_setting.py", line 282, in to_disk
    raise FileExistsError(f"File already exists! {self.path}")
```
and no .nc file of course. However, this type of error is not registered in the log file.

Finally, I have a suggestion. I tried different time settings, shot_lists, and output settings… It would be nice that any of these options in RetrievalSettings() or get_shots_data() are saved in the output.json, so even if the code is lost by the user, another user in the future can reproduce the exact data retrieval.

PR #536, opened 2026-03-30
Prefetch DIII-D plasmas rundb table

### LINKED ISSUE #397: Optimize DIII-D code rundb queries
We should revamp the current DIII-D logic to carry out a single query for all shots, and then storing that info in a short-lived cache (eg `/tmp`), as opposed to having each process query the proper EFIT tree of each shot all the time.

gtrevisan:
DIII-D EFIT tree assignments are not static, but rather dynamic as stored in the `code_rundb.dbo.plasmas` table.

previously, all processes where executing individual queries for each shot.

now we prefetch the whole table with respect to a given runtag -- caching it to disk into the "daily" default temporary folder (`/local-scratch/$USER/disruption-py/YYYY-MM-DD`).

downstream forked processes, and further repeat executions, should be hitting the cache and skipping queries altogether.

suggested reviewers:
- @samc24 hopefully a quick look to tell me whether the implementation makes sense (and to be aware of any possible conflict with your backend abstraction);
- @yumouwei @ZanderKeith as DIII-D users, you should be able to quickly test functionality and usability. 

we can change runtag from the default `DIS` as follows:
```bash
DISPY_EFIT__RUNTAG=DISPY uv run disruption-py -l debug
```


yumouwei:
The code does work. However, do we need to check if the `code_rundb.dbo.plasmas` table has been modified when there's already a cached `rundb_DIS.csv` or `rundb_DISPY.csv` file in the scratch folder?

gtrevisan:
yes, I thought about that but decided against it for simplicity, at least for now.

I'm currently using a "daily" temporary folder, so there is a very slim chance that something cool happens to the table after we cached it but before we rerun a workflow.

in the specific case of our DisruptionEFIT runs, those happen in the wee hours of the morning, so even by our eastern 8am everything is completed and archived.

if someone was running EFIT for DIII-D and then expecting to find the tree right away, then yes, they should remove the cached CSV.

(unrelated to the db, but) if someone was running EFIT for C-MOD they would probably still have to wait overnight in this big mess of servers and trees (archive vs test vs new, etc), or carefully craft their `default_tree_path` lookup env var.

bottom line, let's see how this works for now!

yumouwei (approved):

disruption_py/settings/nickname_setting.py
```diff
@@ -76,6 +98,18 @@ def get_tree_name(self, params: NicknameSettingParams) -> str:
                 return self.tokamak_overrides[params.tokamak](params)
         return self._get_tree_name(params)
 
+    def prefetch_db(self, database: ShotDatabase, tokamak: Tokamak) -> None:
```
samc24, line 101: nit: Concrete no-op methods usually by standards should have an explicit `pass` to portray "intentionally empty" . The abstract methods in this file are fine without it (the @abstractmethod decorator signals intent there), but this one isn't abstract.

disruption_py/settings/nickname_setting.py
```diff
@@ -228,6 +277,81 @@ def __init__(self):
...
+        df = df.set_index("idx")
+        self._d3d_efit_trees = _efit_df_to_dict(df)
+
+        # write cache
+        logger.debug("Caching EFIT '{runtag}' trees: {csv}", runtag=runtag, csv=csv)
+        df.to_csv(tmp)
+        if os.path.exists(csv):
+            logger.warning("Removing stale EFIT cache: {csv}", csv=csv)
+            os.remove(csv)
+        os.link(tmp, csv)
```
samc24, line 354: Minor: os.replace(tmp, csv) is atomic on POSIX and replaces the remove + link two-step. Avoids a small potential race window where another process could interfere between the remove and the link. 
non-blocking but maybe this would be better and preserve the log too if we care abt that 
  ``` 
        if os.path.exists(csv):
           logger.debug("Replacing EFIT cache: {csv}", csv=csv)
        os.replace(tmp, csv)
```

File: disruption_py/settings/nickname_setting.py (hunk 1 above)
Comment (gtrevisan) [reply disruption_py/settings/nickname_setting.py:101]: thanks for the tip! honestly in the recent past I've been pushed by linting tools (eg pylint) to remove useless `pass` statements, as it would be indeed here due to the presence of docstrings. I'll try and put it in, and if it passes the linting workflow I'll push it!

**Comment by samc24** [comment]
LGTM, clean sensible optimization, no conflicts with the backend abstraction, or with potential work on extracting shared fields from Params classes. Two minor nits added but non-blocking so feel free to ignore and merge .

File: disruption_py/settings/nickname_setting.py (hunk 1 above)
Comment (samc24) [reply disruption_py/settings/nickname_setting.py:101]: oooh interesting hahaha didnt know that. fair enough to leave as is then lets not upset the linter lords

**Review by samc24 — APPROVED** [review]

File: disruption_py/settings/nickname_setting.py (hunk 2 above)
Comment (gtrevisan) [reply disruption_py/settings/nickname_setting.py:354]: I went with a hardlink since it should be faster as it does not need to create a new inode, but yes, in both cases there might be a microscopic surface for race conditions.
ideally, though, concurrent processes will fetch and cache the same exact csv file, so I'm not too anxious about that.
in any case, each process that gets here should use and parse what it read fresh from the DB, so only subsequent later processes will use such cached data.
for the minor use case, here, I think we'll be fine. :turtle:

File: disruption_py/settings/nickname_setting.py (hunk 1 above)
Comment (gtrevisan) [reply disruption_py/settings/nickname_setting.py:101]: yep, confirmed: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unnecessary-pass.html

**Comment by gtrevisan** [comment]
ah, for the record, this is the scale and speed of our two EFIT runtags on DIII-D:

```
# efit18-equivalent
[ DEBUG ] Fetched EFIT 'DIS' trees in 0.325s: 17,277 rows, 17,157 unique shots.

# efit21-equivalent
[ DEBUG ] Fetched EFIT 'DISPY' trees in 0.406s: 55,603 rows, 55,603 unique shots.

# cache sizes
372K	rundb_DIS.csv
1.2M	rundb_DISPY.csv
```

so I challenge anyone to break this! :rabbit2:

PR #545, opened 2026-04-03
Implement generic current time computation method

### LINKED ISSUE #223: implement disruption time computation
implement disruption time computation as internal method, as opposed to relying on the external `disruptions` table.

references on MFE WSs:
```
/usr/local/cmod/codes/mhd/magnetics/disruption_database/test_disrupt.pro
/home/granetz/Collaborations/DIII-D/disruption_database/test_for_disruption.m
/home/granetz/Collaborations/EAST/disruption_database/test_for_disruption.m
/home/granetz/Collaborations/KSTAR/disruption_warning_database/test_for_disruption_KSTAR.m
```

yumouwei:
# Implemented changes

- Added `GenericPhysicsMethods.get_current_quench_time(...)` and `GenericUtilMethod.get_end_of_current(...)` based on Bob's original scripts in IDL & MATLAB (see #223).
- The computed `current_quench_time` is returned as a column with the same value for each shot. 

# Testing

- So far, I've run spot checks on the testing shots in CMOD & DIII-D, and a few DIII-D shots in the 201xxx range which are recently used in training the DPRF model for the AI Task Force experiment. All shots except `cmod 1150805014` (see below) agrees with the existing `t_disrupt` from the cached SQL database.

# TODOs

- [x] Figure out how to return the `current_quench_time` in the dataset/dataframe.
- [ ] Implement unit test for this method.
- [ ] Apply this method to MAST & HBT-EP and determine the proper threshold values.
  - [x] Complete implementation for MAST & HBT-EP
  - [ ] Determine MAST thresholds
  - [ ] Determine HBT-EP thresholds

# References

- #223 

# Note on `cmod 1150805014`

`cmod 1150805014` is considered a non-disruptive shot in the SQL database. However, the parameters of this shot do satisfy all of the 6 testing criteria. The shot does look like a proper ramp down disruption with sufficient pre-disruption `ip`, so I believe it should be relabeled as disruptive.

image


yumouwei:
# Adapting the method to MAST & HBT-EP
## MAST

I copied in the thresholds from CMOD and tested them with the testing shots. Right now the method is able to locate the current quench event in the disruptive shots, but it occasionally marks a non-disruptive shot as disruptive (see `mast 29695` below). I tried adjusting some of the thresholds but it didn't help. 

I think the problem is that right now this method has some hard-coded assumptions about the machine's disruption & rampdown timescales (e.g. `L219-222` where it checks for the pre-ip-spike maximum current based on the computed shot duration - 50 ms, or `L232-237` where it searches for the candidate t_disrupt within +-50ms of the shot duration). So far these assumptions have worked out fine across CMOD, DIII-D, & EAST, but they should be generalized and become machine-specific so that we can adapt this method to other machines with either faster or slower timescales. I believe this should be done in the next revamp of this method when we plan to create a disruption dataset of MAST.

https://github.com/MIT-PSFC/disruption-py/blob/08cea0aef8a3a01cda11c05d83c72b72748632ae/disruption_py/machine/generic/physics.py#L219-L222

https://github.com/MIT-PSFC/disruption-py/blob/08cea0aef8a3a01cda11c05d83c72b72748632ae/disruption_py/machine/generic/physics.py#L232-L237

image

image

## HBT-EP

For HBT-EP, the problem with the timescale assumptions is more severe. An HBT shot usually lasts between 5~10 ms. It always ends with a disruption, and the disruption precursor usually occurs within the last 1 ms prior to the current quench. These shorter timescales break every assumption about the timescales of current quench events in the method, which makes the method in its current form incompatible with HBT regardless of the threshold values. Similar to the suggestions for MAST, I believe this should be done in a subsequent revamp after a more throughout study of the disruption dynamics in HBT-EP.

yumouwei:
# Comparison of `current_quench_time` vs `t_disrupt`

- The following figures compare the `current_quench_time` computed using this new physics method with `t_disrupt` previously computed by Bob Granetz using all shots from the `disruption_warning` database.

## 1. "Confusion Matrix"

image
image

## 2. Difference between `current_quench_time` vs `t_disrupt`

- These figures include shots that have both `current_quench_time` and `t_disrupt` (i.e. lower right category in the "confusion matrix")

image
image

## Analysis

- The `get_current_quench_time` physics method is mostly robust for D3D shots. This is expected as the method was translated mainly based on the D3D version of `test_for_disruption.m` and `end_of_current_d3d.m`.
  - I also expect it to work well for EAST (& KSTAR once we incorporate it) since their versions of the MATLAB scripts are nearly identical to the D3D version.
- For C-Mod, right now there are a lot of "false positives" (shots that don't have `t_disrupt` but have `current_quench_time`). I spot checked a couple of these shots and many of them are what I'd consider as "disruptive" based on the descriptions in #223 similar to `cmod 1150805014` shown above. One exception I found so far is `cmod 1120502001` shown below.
- Upon further investigation, I found out that there are actually a couple differences between `test_disrupt.pro` (for CMOD) and the other MATLAB scripts：
  - C-Mod uses different search windows to get the intermediate variables like `ip_max`, `ip_final` and `candidate_di_dt`. This is the exact same timescale problem I mentioned earlier about MAST and HBT-EP.
  - C-Mod uses a different method to find the pre-CQ Ip (`candidate_ip0`) by backtracking from `candidate_t_disrupt` until `dI/dt` crosses a threshold value, as opposed to computing the mean Ip between a window based on the end of current.

Based on these analysis I'd expect this method to take more time to complete. Alternatively we can merge it "as-is", keep everything that uses `t_disrupt` intact, and implement the revamps before we decouple with SQL DBs.


## Additional figure

- `cmod 1120502001` (should be labeled as non-disruptive)

image

- `cmod 1140522013` (`Ip0 < 100kA` so should be considered as non-disruptive)

image

gtrevisan:
I ran the new `disruption-errors` script for C-MOD and got:
```
disruption-errors.sh plasmas -m get_current_quench_time

      4  get_current_quench_time: UnboundLocalError("cannot access local variable 'ip' where it is not associated with a value")
     39  get_current_quench_time: ValueError('max() iterable argument is empty')

[ ERROR ] #1120918022 | get_current_quench_time: UnboundLocalError("cannot access local variable 'ip' where it is not associated with a value")
[ ERROR ] #1050505004 | get_current_quench_time: ValueError('max() iterable argument is empty')

1050505004 1120918022
```

gtrevisan:
I ran the new `disruption-errors` script for DIII-D and got:
```
disruption-errors.sh plasmas -m get_current_quench_time

     47  get_current_quench_time: ValueError('attempt to get argmin of an empty sequence')
    443  Failed retrieval! KeyError('param')
   8545  get_current_quench_time: ValueError('max() iterable argument is empty')

[CRITICAL] #70581 | Failed retrieval! KeyError('param')
[ ERROR ] #78257 | get_current_quench_time: ValueError('attempt to get argmin of an empty sequence')
[ ERROR ] #70415 | get_current_quench_time: ValueError('max() iterable argument is empty')

70415 70581 78257
```
(actually, I ran with 1 cpu for roughly 65k out of 75k shots, then it was grinding almost to a halt for some MDSplus reason.)

yumouwei:
Analysis of the 5 identified problematic shots

| Shot            | Identified cause                                                                                                                        | Implemented change                                                                                             |
|-----------------|-----------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------|
| cmod 1050505004 | `get_end_of_current` calculates duration=0.0914 which does not satisfy the bona fide plasma condition by Bob, then returns duration=0.  | Add check for duration==0 (i.e. not bona fide plasma) in `get_current_quench_time`                             |
| cmod 1120918022 | `ip` = array of nan (no ip signal)                                                                                                              | Remove the `try-except` block around `get_data_with_dims` and let the framework catch MDSplus errors           |
| d3d 70415       | ip was not digitized to the end of the discharge. `get_end_of_current` marks this by setting `duration = -duration`                     | Check for negative `duration` and raise `CalculationError`                                                     |
| d3d 70581       | Typo in `ValueError`                                                                                                                    | Corrected typo, change it to `CalculationError`                                                                |
| d3d 78257       | `ip_upright[time_indices]` has only 1 element, thus `didt_upright = []`.                                                                | Check for `len(didt_upright)==0` as well as other outputs of `np.where`, raise `CalculationError` accordingly. |

zapatace:
I reviewed 300 random shots from MAST. It works reasonably well. 

I found some corner cases in which the method fails returning all NaNs. 18 cases out of 300 so 6%.

Few examples:

image

image

image

image

image

The list of pathological shots is 16553, 19792, 23730, 25287, 25459, 25735, 26606, 26612, 26617, 26842, 28219, 28232, 29077, 29136, 29146, 29645, 30011, 30278

I'll continue investigating...

yumouwei:
Only 18 bad cases out of 300 looks promising, so I'd recommend merging as is and adjust them once we want to create the MAST disruption dataset.

Of the example shot:
- 16553 looks like a ramp down disruption and I'm surprised it didn't get caught
- 19792 can be argued as a good ramp down although it does have a "minor-ish disruption" at 0.4 s. I don't know how we would classify this shot in the current paradigm, and I don't think the current quench time method is designed to catch this.
- 23730 "might" be a mitigation shot. I've seen similar ip trace shape from 0.4-0.5s on d3d mitigation shots so I guess it could be something similar.
- 25459 I have no idea what's going on. Looks like it's either a vacuum power supply testing shot or the sensor was just broken.
- 25753 should be a proper ramp up disruption. I'm also surprised that the method didn't catch it.

gtrevisan:
- #555
- #554

disruption_py/machine/cmod/config.toml
```diff
@@ -37,6 +37,10 @@ description = "Magnetic chi^2 (EFIT)."
 units = "dimensionless"
 validity = [0.1, 40]
 
+[cmod.physics.attributes.current_quench_time]
```
gtrevisan, line 40: imas?

yumouwei: I'd leave this empty for now. IMAS does have this very weird `summary/disruption/time` which is defined as 

> "Time of the disruption = min(t_TQ, t_VDE), where t_TQ and t_VDE are the times at which the Thermal Quench (TQ) and the Vertical Displacement Event (VDE) are detected by their respective detectors."

 It does not have a structure for the current quench time.

gtrevisan: what about `disruption` -- not `summary`?

- https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/disruption.html#disruption-time

yumouwei: That one has the type `FLT_1D` so I assume it's just a generic time array. The `disruption` category also has structures like `code` & `halo_currents` which makes me think it's designed to store results from some kind of current quench simulation.

zapatace:
I reviewed the 300 shots I had for reviewing the QC method (https://github.com/MIT-PSFC/disruption-py/pull/545#issuecomment-4389520958). If I discard rampdowns that should not detect, and partial disruptions. Error is ~1%, pretty acceptable. I will approve the PR. 

I also agree on the IMAS issue (https://github.com/MIT-PSFC/disruption-py/pull/545#discussion_r3268443670). They don't have the same definition as our QC method, we should not populate the IDS then.

zapatace (approved):
QC method works in MAST

gtrevisan (approved):

PR #547, opened 2026-04-09
Fix EPSOFF logic for post-198183 shots

### LINKED ISSUE #506: Fix EPSOFF check logic in D3D physics method
According to Kurt Holtrop: 

> The wiring was changed during the 2023/2024 vent so between the last plasma shot of 2023 and the first power supply test shot of 2024. The last plasma shot on 6/29/23 was 196645. (pre-change) The first low power supply test shot of 2024 was 198183 on 4/24/24. (post-change)

Affected physics methods:

- get_ip_parameters
- get_rt_ip_parameters

yumouwei:
# Implemented changes

- Consolidate the identical check on `epsoff` in `get_ip_parameters` and `get_rt_ip_parameters` into a single `get_power_supply_railed` physics method.
- Implement the wiring switch described in Issue #506 in `get_power_supply_railed`

# Summary of behavior change

| shot_id | before (`dev`)  | after (`pr547`) |
| - | - | - |
| < 198183 | abs(epsoff) > 0.5 = bad | no change |
| >= 198183 | abs(epsoff) > 0.5 = bad | abs(epsoff) < 0.5 = bad |

- bad: E-coil power supply have railed and PCS feedback control of Ip is not being applied in those time points. Therefore `ip_error` and `ip_error_rt` are undefined for those times.

# Example plots

d3d_shot_201927_before
d3d_shot_201927_after

- 201927: post wiring switch, behavior of epsoff was verified with d3d control group.

# References

- #506


disruption_py/machine/d3d/physics.py
```diff
@@ -610,6 +610,7 @@ def get_ip_parameters(params: PhysicsMethodParams):
         -------
         - original source: [get_Ip_parameters.m](https://github.com/MIT-PSFC/disruption-py/blob
         /matlab/DIII-D/get_Ip_parameters.m)
+        - issues: #[506](https://github.com/MIT-PSFC/disruption-py/issues/506)
```
gtrevisan, line 661: you might want to add this PR, as well.

disruption_py/machine/d3d/physics.py
```diff
@@ -684,7 +685,11 @@ def get_ip_parameters(params: PhysicsMethodParams):
             # on the last time sample
             t_epsoff += 0.001
             epsoff = interp1(t_epsoff, epsoff, params.times, "linear")
-            railed_indices = np.where(np.abs(epsoff) > 0.5)
+            if params.shot_id >= 198183:
```
gtrevisan: since we are using this in two places, let's choose a good name and make it a configuration parameter.

disruption_py/machine/d3d/physics.py
```diff
@@ -734,6 +739,7 @@ def get_rt_ip_parameters(params: PhysicsMethodParams):
         - original source: [get_Ip_parameters_RT.m](https://github.com/MIT-PSFC/disruption-py
         /blob/matlab/DIII-D/get_Ip_parameters_RT.m)
         - pull requests: #[254](https://github.com/MIT-PSFC/disruption-py/pull/254)
+        - issues: #[506](https://github.com/MIT-PSFC/disruption-py/issues/506)
```
gtrevisan, line 781: you might want to add this PR, as well.

gtrevisan (commented):
what's the final effect of this change?
are you able to share a couple of plots for some shot?

also, have you tried executing full tests to see whether we need to mark anything as an expected failure?

disruption_py/machine/d3d/physics.py
```diff
@@ -570,6 +570,60 @@ def get_rt_density_parameters(params: PhysicsMethodParams):
...
+            t_epsoff += 0.001
+            epsoff = interp1(t_epsoff, epsoff, params.times, "linear")
+            power_supply_railed = np.zeros(len(params.times))
+            if params.shot_id >= 198183:
+                # Post 2023/2024 wiring change: high=good, 0=bad
+                (railed_indices,) = np.where(np.abs(epsoff) < 0.5)
+            else:
+                (railed_indices,) = np.where(np.abs(epsoff) > 0.5)
+            power_supply_railed[railed_indices] = 1
+        except mdsExceptions.MdsException as e:
```
gtrevisan: are you sure we should catch this?
we could let the framework do it.

gtrevisan (commented):
- do you want more in-depth review on this? I assume you don't.
- I'll tag Copilot to see if it comes up with any suggestion.
- my minor comment relates to error handling, again... was there a reason for the try/except block?
- have you run this against eg the `disruption_warning` shotlist to see whether `disruption-errors` catches anything interesting?

disruption_py/machine/d3d/physics.py
yumouwei: I thought about this, but I think it'd be better to only put threshold values used by generic methods in the config files.

gtrevisan: you could place it in a new lower ad-hoc section, not at the top generic section, eg:
```toml
[d3d.physics.get_power_supply_railed]
```
or something similar -- this would parameterize it without putting it under everybody's eyes.

gtrevisan:
```
./scripts/disruption-errors.sh disruption_warning \
-m get_power_supply_railed -m get_ip_parameters -m get_rt_ip_parameters
```
... no errors? :medal_sports: !

disruption_py/machine/d3d/physics.py
gtrevisan: I'll drop this for now.

disruption_py/machine/d3d/physics.py
```diff
@@ -733,7 +775,9 @@ def get_rt_ip_parameters(params: PhysicsMethodParams):
         -------
         - original source: [get_Ip_parameters_RT.m](https://github.com/MIT-PSFC/disruption-py
         /blob/matlab/DIII-D/get_Ip_parameters_RT.m)
-        - pull requests: #[254](https://github.com/MIT-PSFC/disruption-py/pull/254)
+        - pull requests: #[254](https://github.com/MIT-PSFC/disruption-
+        py/pull/254), #[547](https://github.com/MIT-PSFC/disruption-py/pull/547)
+        - issues: #[506](https://github.com/MIT-PSFC/disruption-py/issues/506)
```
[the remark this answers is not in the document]
yumouwei: These line breaks do work with mkdocs

disruption_py/machine/d3d/physics.py
```diff
@@ -570,6 +570,54 @@ def get_rt_density_parameters(params: PhysicsMethodParams):
...
+        # Avoid problem with simultaneity of epsoff being triggered exactly
+        # on the last time sample
+        t_epsoff += 0.001
+        epsoff = interp1(t_epsoff, epsoff, params.times, "linear")
+        power_supply_railed = np.zeros(len(params.times))
+        if params.shot_id >= 198183:
+            # Post 2023/2024 wiring change: high=good, 0=bad
+            (railed_indices,) = np.where(np.abs(epsoff) < 0.5)
+        else:
+            (railed_indices,) = np.where(np.abs(epsoff) > 0.5)
```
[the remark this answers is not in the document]
yumouwei: The last shot in the `disruption_warning` table is 177061 which is before the logic switch, so we won't be able to test this with our current `test_against_cache` approach.

disruption_py/machine/generic/physics.py
```diff
@@ -70,9 +70,10 @@ def get_time_domain(params: PhysicsMethodParams):
             signals["ip_prog"] = ip_parameters["ip_prog"]
         elif params.tokamak == Tokamak.D3D:
             ip_parameters = D3DPhysicsMethods.get_ip_parameters(params=params)
+            psr_parameter = D3DPhysicsMethods.get_power_supply_railed(params=params)
             signals["dipprog_dt"] = ip_parameters["dipprog_dt"]
             signals["ip_prog"] = ip_parameters["ip_prog"]
-            signals["power_supply_railed"] = ip_parameters["power_supply_railed"]
+            signals["power_supply_railed"] = psr_parameter["power_supply_railed"]
```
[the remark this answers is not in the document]
yumouwei: This does work as intended:

- WIth the previous logic, an `mdsException` will make `get_ip_parameters` to return`power_supply_railed = np.array(all nans)`, which then trigger L86-89 and returns `time_domain` as `[np.nan]`
https://github.com/MIT-PSFC/disruption-py/blob/2085bd0ba63f48135140b71a63aa49ceb98d51a2/disruption_py/machine/generic/physics.py#L86-L89

- With the modified logic, the `mdsException` will trigger an error in `get_power_supply_railed` and a `CalculationError` warning for `get_time_domain`, and then returns `time_domain` as `[np.nan]`

disruption_py/machine/d3d/physics.py
```diff
@@ -570,6 +570,54 @@ def get_rt_density_parameters(params: PhysicsMethodParams):
...
+        - issues: #[506](https://github.com/MIT-PSFC/disruption-py/issues/506)
+        """
+        epsoff, t_epsoff = params.data_conn.get_data_with_dims(
+            f"ptdata('epsoff', {params.shot_id})"
+        )
+        t_epsoff = t_epsoff / 1.0e3  # [ms] -> [s]
+        # Avoid problem with simultaneity of epsoff being triggered exactly
+        # on the last time sample
+        t_epsoff += 0.001
+        epsoff = interp1(t_epsoff, epsoff, params.times, "linear")
```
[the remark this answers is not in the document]
yumouwei: This should already be handled by the framework

disruption_py/machine/d3d/physics.py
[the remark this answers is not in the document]
yumouwei: This is actually an interesting question. So far I haven't seen a case where `epsoff` contains individual `nan` in an otherwise valid array. If such cases exist, I do not know if we should interpret these `nans` as "railed" (i.e. bad points/should be filtered out) or just "missing data" (i.e. fine to keep). I'd suggest keeping the logic as is and modify it if this problem arises down the line.

gtrevisan (approved):

PR #548, opened 2026-04-13
Update workflow figure

yumouwei:
- Moved MAST from "Working in Progress" to "Supported Devices" category


gtrevisan (changes requested):
sorry, I never noticed that near CTH you write "Working in Progress" while it should be probably "Work in Progress".

also, you need to update your second box from MDSplus trees to MDSplus trees or Xarray store.

feel free to force push the branch with the new workflow! thanks.

yumouwei:
Made the requested changes. I don't want to force-push in case I break anything.

gtrevisan (approved):

PR #551, opened 2026-04-15
Add disruption-errors utility workflow

gtrevisan:
this script parses a `disruption-py` log file and collects unexpected errors.

- first, it runs `disruption-py` by passing on any command-line argument, but can optionally digest a previous log file, instead.
- then, it parses the log file and collects all errors and critical messages, excluding those that match the patterns in the sibling .ignore file.
- finally, it prints a useful summary, consisting of:

  - a :red_circle: list of unique errors, together with their relative frequency;
  - a :yellow_circle: list of error examples, one per unique error;
  - a :green_circle: list of shot numbers, one per unique error.

usage examples:
```
disruption-errors.sh -m get_ip_parameters disruption_warning
disruption-errors.sh /path/to/run/output.log
ls /path/to/run/*log | xargs -n1 disruption-errors.sh
```

### DIFF
--- a/scripts/disruption-errors.ignore
+++ b/scripts/disruption-errors.ignore
@@ -1,0 +1,7 @@
+MDSPLUS-E-ERROR
+MDSplusERROR
+NotImplementedError
+parameters have data
+Pathological timebase
+TreeFOPENR
+TreeNODATA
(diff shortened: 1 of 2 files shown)


yumouwei (approved):
I tested this on cmod & d3d and it worked

gtrevisan:
in the future we should probably add another logger to output JSON-serialized data, but this suffices for now.

PR #552, opened 2026-04-21
Fix double-cached physics methods

gtrevisan:
- do not double-cache physics methods, as they are already cached by their decorator,
- also, cache a couple of minor private methods.


gtrevisan:
@copilot are there other private methods that require caching?
anything that gets called at least twice would require the @cache_method decorator.

PR #553, opened 2026-04-27
Adding p_oh method for MAST

### LINKED ISSUE #543: Revamp ohmic power methods
assignments:
- @AlexSaperstein to lead the effort,
- @hwietfeldt and @granetz for support and improved calculation with specific focus on radiative collapses,
- @yumouwei for awareness and help with evaluating whether these methods could become generic methods.

AlexSaperstein:
Adding Ohmic heating power feature/method for the MAST tokamak.

The method used is based on that used for the same calculation for C-Mod, which uses Ip, li(3), and vloop.
- MAST has multiple vloop calculations available. The _dynamics_ vloop was used in this calculation over the _static_ calculations, as it allows for the motion of the LCFS, making it strictly more accurate.
- This method, that C-Mod also currently uses, does not consider the Ip*dL/dt contribution to V_inductive, which gets subtracted from Vloop to reconstruct V_resistive. This implies that the p_oh calculation may be less accurate during periods of fast li(3) changes, like during temperature profile collapses. A correction to p_oh that properly accounts for this term is in development.


disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,51 @@ def get_dalpha(params: PhysicsMethodParams):
...
+            The parameters containing the Xarray connection, shot id and more.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the total Ohmic heating power (`p_oh`).
+        """
+
+        # load relevant parameters
+        conn: XarrayConnection = params.mds_conn
```
gtrevisan: this is now `data_conn`, without type hints:
```suggestion
        conn = params.data_conn
```

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,51 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        r0 = conn.get_data(params.shot_id, "equilibrium/magnetic_axis_r")
```
gtrevisan: I don't think you need the shot id anymore:
```suggestion
        r0 = conn.get_data("equilibrium/magnetic_axis_r")
```

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,51 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        li = conn.get_data(params.shot_id, "equilibrium/li")
+        v_loop = conn.get_data(params.shot_id, "equilibrium/vloop_dynamic")
+        ip = conn.get_data(params.shot_id, "summary/ip")
+        summary_time = conn.get_data(params.shot_id, "summary/time")
+        equilibrium_time = conn.get_data(params.shot_id, "equilibrium/time")
+        dip_dt = np.gradient(ip, summary_time)
```
gtrevisan, line 357: you might want to add a newline and a new comment, eg:
```suggestion

        # compute derived quantities
        dip_dt = np.gradient(ip, summary_time)
```

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,51 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        dip_smoothed = causal_boxcar_smooth(dip_dt, 6)
+        inductance = 4.0 * np.pi * 1.0e-7 * r0 * li / 2.0
+
+        # interolate to consistent time-base
```
gtrevisan: typo:
```suggestion
        # interpolate to consistent time-base
```

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,51 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        # calculate p_oh
+        v_inductive = inductance * dip_smoothed
+        v_resistive = v_loop - v_inductive
+        p_ohm = ip_interp * v_resistive
+        
+        # Set negative p_ohm values to 0
+        (indices,) = np.where(p_ohm < 0)
+        p_ohm[indices] = 0
+
+        return {"p_oh": p_ohm}
```
gtrevisan: you should add a proper newline at the end of the file.

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,51 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        # interolate to consistent time-base
+        v_loop = interp1(equilibrium_time, v_loop, params.times)
+        inductance = interp1(equilibrium_time, inductance, params.times)
+        dip_smoothed = interp1(summary_time, dip_smoothed, params.times)
+        ip_interp = interp1(summary_time, ip, params.times)
+
+        # calculate p_oh
+        v_inductive = inductance * dip_smoothed
+        v_resistive = v_loop - v_inductive
+        p_ohm = ip_interp * v_resistive
```
gtrevisan: at this point to avoid confusion you might as well call it:
```suggestion
        p_oh = ip_interp * v_resistive
```
I don't remember how we call it for other machines, but it's probably easier to use the same name for the internal parameter and the external feature.

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,51 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        
+        # Set negative p_ohm values to 0
+        (indices,) = np.where(p_ohm < 0)
+        p_ohm[indices] = 0
```
gtrevisan: what about this?
```suggestion
        p_ohm = np.clip(p_ohm, 0, None)
```

gtrevisan (changes requested):
great, thank you, Alex!

it looks like you branched off an old version of dev, though, so you might want to rebase off the current one -- probably better, just reapply your code to the current dev.

gtrevisan:
furthermore, please try and add the appropriate attributes in:
- https://github.com/MIT-PSFC/disruption-py/blob/dev/disruption_py/machine/mast/config.toml

finally, I'll try running at scale for you in order to detect any outliers that might require more function logic... I'll report back here.

gtrevisan:
after ~1,000 shots I've only got errors like this one:
```
[ ERROR ] #11812 | get_ohmic_parameters: ValueError('x and y arrays must be equal in length along interpolation axis.')
```
the debug log does not show sizes, but I imagine you'll have some NaNs or 0- or 1-long arrays in there.

disruption_py/machine/mast/physics.py
AlexSaperstein: sounds good. I was just basing this off of the format I saw for other methods in this file

disruption_py/machine/mast/physics.py
AlexSaperstein: i used the same syntax that is used for cmod, but this does look cleaner

AlexSaperstein:
> it looks like you branched off an old version of dev, though, so you might want to rebase off the current one -- probably better, just reapply your code to the current dev.

Is there a way to reapply to the current dev while retaining this PR?

gtrevisan:
I'd remove your local branch, branch off again from the latest dev, then copy back the file and continue editing.
finally, push-force a new commit to this branch.

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +337,53 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        Returns
+        -------
+        dict
+            A dictionary containing the total Ohmic heating power (`p_oh`).
+        """
+
+        # load relevant parameters
+        conn = params.data_conn
+        r0 = conn.get_data("equilibrium/magnetic_axis_r")
+        li = conn.get_data(params.shot_id, "equilibrium/li")
```
gtrevisan: just to be clear, I think the shot id needs to be dropped from _all_ `get_data` calls!

AlexSaperstein:
> I'd remove your local branch, branch off again from the latest dev, then copy back the file and continue editing. finally, push-force a new commit to this branch.

I just did this, and fixed the other stuff as well

disruption_py/machine/mast/physics.py
```diff
@@ -216,7 +216,7 @@ def _get_densities(times, n_e, t_n, ip, t_ip, a_minor, t_a):
         Calculate electron density, its time derivative, and the Greenwald fraction.
 
         Parameters
-        ----------
+        ----------causal_boxcar_smooth
```
gtrevisan: hello there

gtrevisan:
> furthermore, please try and add the appropriate attributes in:
> 
>     * https://github.com/MIT-PSFC/disruption-py/blob/dev/disruption_py/machine/mast/config.toml
> 
> 
> finally, I'll try running at scale for you in order to detect any outliers that might require more function logic... I'll report back here.

&

> after ~1,000 shots I've only got errors like this one:
> 
> ```
> [ ERROR ] #11812 | get_ohmic_parameters: ValueError('x and y arrays must be equal in length along interpolation axis.')
> ```
> 
> the debug log does not show sizes, but I imagine you'll have some NaNs or 0- or 1-long arrays in there.

disruption_py/machine/mast/physics.py
AlexSaperstein: haha good catch

AlexSaperstein:
> after ~1,000 shots I've only got errors like this one:
> 
> [ ERROR ] #11812 | get_ohmic_parameters: ValueError('x and y arrays must be equal in length along interpolation axis.')
> 
> the debug log does not show sizes, but I imagine you'll have some NaNs or 0- or 1-long arrays in there.

Checking this shot specifically, it does look like one of the signals (vloop) is only 1-long.

What should we do for these instances? How often is this error showing up?

gtrevisan:
fixing that error requires further thought -- it might be our xarray backend returning NaNs instead of raising KeyError -- so all good for now!

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +331,52 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        li = conn.get_data("equilibrium/li")
+        v_loop = conn.get_data("equilibrium/vloop_dynamic")
+        ip = conn.get_data("summary/ip")
+        summary_time = conn.get_data("summary/time")
+        equilibrium_time = conn.get_data("equilibrium/time")
+
+        # compute derived quantities
+        dip_dt = np.gradient(ip, summary_time)
+        dip_smoothed = causal_boxcar_smooth(dip_dt, 6)
+        inductance = 4.0 * np.pi * 1.0e-7 * r0 * li / 2.0
```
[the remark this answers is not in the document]
gtrevisan: will deal this in a separate PR.

disruption_py/machine/mast/physics.py
```diff
@@ -331,3 +331,52 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        r0 = conn.get_data("equilibrium/magnetic_axis_r")
+        li = conn.get_data("equilibrium/li")
+        v_loop = conn.get_data("equilibrium/vloop_dynamic")
+        ip = conn.get_data("summary/ip")
+        summary_time = conn.get_data("summary/time")
+        equilibrium_time = conn.get_data("equilibrium/time")
+
+        # compute derived quantities
+        dip_dt = np.gradient(ip, summary_time)
+        dip_smoothed = causal_boxcar_smooth(dip_dt, 6)
```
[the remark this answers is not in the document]
gtrevisan: @AlexSaperstein I'd probably check the length as suggested, and then change logic or raise a `CalculationError` -- your choice.

disruption_py/machine/mast/physics.py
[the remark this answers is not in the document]
AlexSaperstein: @gtrevisan looks like co-pilot caught the issue as well

gtrevisan (approved):
LGTM. waiting for @yumouwei's review... or maybe @zapatace's?

yumouwei:
@AlexSaperstein can you put a few bullet points in the PR description to describe how you implemented this method? e.g. you can mention the distinction between static vs dynamic vloop, and incorporating only the `L*dI/dt` term in the calculation. Other than that this PR looks good and should be ready to go.

yumouwei (approved):

AlexSaperstein:
The p_oh signals were coming out very noisy. I found that this was because the dip_smoothed signal was still very noisy as well, despite having the same smoothing settings as those used for C-Mod. The reason why dip is noisier for MAST than C-Mod is that the time-base for Ip that the MAST database provides is much slower than what we have for C-Mod. So when we smooth over the same 6 points, the MAST signal remains noisier.

To minimize this noise, while also keeping delays acceptable, the smoothing window for the MAST calc of p_oh was updated from 6-->30. Comparing the 6-point and 30-point smoothed p_oh signals (below), this larger window increases the delay to ~10ms. 

image

image

This ~10ms delay should be acceptable for the time being, and is certainly more beneficial than having too noisy of a p_oh signal.

gtrevisan (commented):
will we need to tweak this parameter further?
maybe we should make it a configuration.
@yumouwei your call!

in any case, @AlexSaperstein, I'd imagine you might want to update your `len` check to the same number, right?

AlexSaperstein:
> will we need to tweak this parameter further? maybe we should make it a configuration. @yumouwei your call!
> 
> in any case, @AlexSaperstein, I'd imagine you might want to update your `len` check to the same number, right?

It is something we could tweak in the future, but it's not something that I think we have to do

yumouwei:
I agree with Alex. LTGM and see if there's any problem with the computed p_oh, then we can adjust the filtering accordingly.

gtrevisan:
- #555
- #554

AlexSaperstein:
> apologies for the conflicts, but please merge `dev` so as to pull in changes from:
> 
>     * [Simplify calls to data getters #555](https://github.com/MIT-PSFC/disruption-py/pull/555)
> 
>     * [Introduce CustomErrors and DataErrors #554](https://github.com/MIT-PSFC/disruption-py/pull/554)

It's all ready

PR #561, opened 2026-05-28
Fix conflicting attrs during concat

gtrevisan:
I realized that the upcoming ~py3.14~ spawn/forkserver issues will affect my LRU-cached `get_metadata` method.
~with the new python version~ a slightly different time will be returned for each process and therefore xarray will flag attribute conflicts.

the LRU cache was supposed to prevent this, as the metadata should be resolved (and its cache should be primed) as a first thing by the logger:
https://github.com/MIT-PSFC/disruption-py/blob/0063f4be23d5aa8d79e6e2c4e98293f02f629e5c/disruption_py/settings/log_settings.py#L157

as a dirty fix, we can drop conflicts while concatenating and reapply them on the final product, here:
https://github.com/MIT-PSFC/disruption-py/blob/0063f4be23d5aa8d79e6e2c4e98293f02f629e5c/disruption_py/settings/output_setting.py#L328

@samc24 feel free to dig around whether there is a better way to solve this!
or LMK if you cannot replicate it -- any multiproc multishot workflow should do, including the testing ones.

eg:

```
ERROR tests/test_output_setting.py::test_output_exists
xarray.structure.merge.MergeError: combine_attrs='no_conflicts',
but some values are not the same. Merging
{ [...] 'time': '2026-05-28T15:23:11.087869', [...] }
with
{ [...] 'time': '2026-05-28T15:23:11.430584', [...] }
```

### DIFF
--- a/disruption_py/settings/output_setting.py
+++ b/disruption_py/settings/output_setting.py
@@ -18,7 +18,7 @@ import pandas as pd
 import xarray as xr
 from loguru import logger
 
-from disruption_py.core.utils.misc import get_temporary_folder
+from disruption_py.core.utils.misc import get_metadata, get_temporary_folder
 from disruption_py.machine.tokamak import Tokamak
 
 
@@ -325,7 +325,8 @@ class DatasetOutputSetting(SingleOutputSetting):
             logger.critical("Nothing to concatenate!")
             return xr.Dataset()
 
-        ds = xr.concat(self.results.values(), dim="idx", combine_attrs="no_conflicts")
+        ds = xr.concat(self.results.values(), dim="idx", combine_attrs="drop_conflicts")
+        ds.attrs.update(get_metadata())
         if "shot" not in ds.coords or "time" not in ds.coords:
             return ds
 
--- a/tests/test_output_setting.py
+++ b/tests/test_output_setting.py
@@ -95,7 +95,7 @@ def test_output_exists(fresh_data, test_folder_m):
 
     # format equivalence
     for ds_list in [dict_out.values(), [dt.to_dataset() for dt in dt_out.values()]]:
-        xr.testing.assert_identical(
+        xr.testing.assert_equal(
             ds_out, xr.concat(ds_list, dim="idx").sortby(["shot", "time"])
         )
     pd.testing.assert_frame_equal(df_out, ds_out.to_dataframe()[df_out.columns])


samc24:
Dug into this. Found the root cause, and there's a fix at the same line count that makes the conflict impossible instead of tolerated — pushed it as `sameerc/fix-concat-attrs` off `dev` so you can compare.

**Root cause.** The LRU cache *is* primed first thing by the logger — but only in the main process. The stamp is applied in the worker (`populate_shot` in `runner.py`), and under spawn/forkserver each worker is a fresh interpreter that re-primes its own cache with its own microsecond. Same spawn-doesn't-inherit-parent-state family as #546 and #560. It also explains the long-standing `test_output_exists` flakiness on macOS: `pool.imap` sometimes hands both shots to one worker (same cached time, passes), sometimes splits them (MergeError).

**Fix.** Move the stamp from the worker to the main process, in the `pool.imap` loop in `workflow.py`, right as each shot comes back. The main-process cache is primed once at `setup_logging()`, so every shot carries bit-identical attrs regardless of start method, and concat never sees a conflict. Two lines moved, no new behavior at concat time.

I know Copilot raised a few adjacent points and I agree none of them are a big deal while the hack works — but with the stamp in the main process they all just go away: `no_conflicts` stays as a tripwire, `assert_identical` stays untouched, per-shot outputs (`dict` folder, DataTree children) carry the same run metadata as the final product, and `output_setting.py` never needs to import `get_metadata`. I put the stamp in the workflow loop rather than `output_shot` — single choke point, and the output layer stays metadata-agnostic.

The final concatenated Dataset has identical attrs either way (the reapplied `get_metadata()` is the same primed dict), so the end product doesn't change. Semantics-wise `time` now uniformly means "workflow start", consistent with the rest of the metadata (commit, version, user) which already describes the run, not the shot.

One thing I checked since it's the load-bearing claim: the drop-and-reapply approach genuinely needs the `assert_identical` → `assert_equal` downgrade. With drop-and-reapply, the per-shot dict outputs keep their differing worker `time`, so the cross-format concat in `test_output_exists` can never be `identical` to the final product again — confirmed it fails with `assert_identical` kept. The main-process stamp keeps the strict check.

Verified locally on MAST (macOS, spawn, 2 fast shots / 2 processes): dev reproduces the exact MergeError (attrs differ only in `time`, ~ms apart), the fix passes `test_output_exists` with `assert_identical` untouched, and `make test-fast` is green.

Happy to retarget onto your branch if you'd rather keep it on this PR — otherwise feel free to pull `sameerc/fix-concat-attrs`. Wanted it in front of you before you head out.

gtrevisan:
awesome! I think you hit the nail on the head. let's close this in favor of your better fix.

PR #562, opened 2026-05-29
Add three new C-Mod physics methods with H-mode related variables

zapatace:
This PR adds three new physics methods to `disruption_py/machine/cmod/physics.py`, along with their attribute definitions (description, IMAS reference, units, validity limits) in `disruption_py/machine/cmod/config.toml`. Following review it also adds `get_btor`, which consolidates the toroidal-field fetch that `get_h98`, `get_lh_power_threshold` and `get_n_equal_1_amplitude` were each doing on their own, and which takes over the pre-existing `bt` column.

| Column | Method | Units | Validity limits |
|---|---|---|---|
| `h98` | `get_h98` | dimensionless | `[0.05, 1.75]` |
| `h_alpha` | `get_h_alpha` | W/(m²·sr) | `[0.0, 161.920898]` |
| `lh_power_threshold` | `get_lh_power_threshold` | W | `[0.0, 1.0e7]` |
| `bt` (pre-existing column) | `get_btor` (new method) | T | — |

I used the following shot as reference example (1100204004).
image

In this example, there are 6 alternating LH transitions followed by HL transition. Notice how H98 approach one, while H-alpha diminish just after Pinput -Prad crosses P_thr from Martin scaling (`lh_power_threshold`), although not always.

## Validation / error analysis
I run the workflow on scale using the disruption warning table, with the final code and the four methods in a single pass (`get_h98`, `get_h_alpha`, `get_lh_power_threshold`, `get_btor`; 4 processes): **10,429/10,435 shots (99.94%) in 6m 52s (0.040 s/shot)**, 940,543 time slices, with finite values on 99.99% (`bt`), 99.8% (`lh_power_threshold`), 98.9% (`h_alpha`) and 97.9% (`h98`) of them. I also run the disruption-errors script on that log and it comes back clean (exit 0, no `ERROR` lines at all): the 6 missing shots are `CRITICAL` setup failures (pathological timebase, missing `efit18` tree/node) raised before any physics method runs.

The new methods never raise an unexpected exception. Their only failures are missing or inconsistent MDSplus inputs, which the framework logs as `WARNING` and fills with NaNs:

- `h_alpha`: 102 shots with no `\spectroscopy::ha_2_bright` data;
- `h98`: 149 shots, all coming from its inputs (`pcs` tree 57, `\twopi_diode` 62, `\lh::top.results:netpow` 22, `\rf_power_net` 3, `magnetics` tree 1, density length mismatch 4);
- `lh_power_threshold`: 5 shots and `bt`: 1 shot (missing `magnetics` tree, density length mismatch).

Next, we have the definition and value distributions for the 3 methods on the disruption warning table.

### `get_h_alpha` → `h_alpha`

H-alpha line emission intensity, useful as a marker of ELMs, radiative events, and confinement-regime transitions. Reads `\spectroscopy::ha_2_bright` [mW/(cm²·sr)] and interpolates onto `params.times`, converting to SI brightness units `W/(m²·sr)` (×10). If the signal is unavailable, the framework catches the exception and fills the column with NaNs. Be adviced that for ELM detection the native signal time base is necessary to avoid losing fast transient events.

image

Multimodal distribution with median ≈ 14 W/(m²·sr), IQR ≈ [5.7, 32]. Some shots show small negative values (min ≈ −104), and there are discrete spikes/saturation lines at `80.523056` and `161.920898` — the latter is used as the upper validity bound.

Negative and saturated shots were inspected, here you have two examples

image
image

I checked with Bob Granetz, and all of this appears to be a unknown diagnostic errors (like wrong offset or maybe some gain adjust that caused the distortion).

### `get_h98` → `h98`

H98 energy-confinement enhancement factor, `h98 = τ_E / τ_98`, where the measured confinement time is `τ_E = W_mhd / (P_input − dW_mhd/dt)` and `τ_98` is the ITER IPB98(y,2) scaling (eq. 20, ITER Physics Basis Ch. 2), evaluated with atomic mass A = 2. Non-physical non-positive values are clipped to 0.

image

Skewed Gaussian peaked around ~0.45 with a long tail toward 1.0 and a large spike at 0 (clipped non-positive values). Raw stats: median ≈ 0.45, IQR ≈ [0.32, 0.57]; `mean`/`max` are `inf` because `P_input − dW_mhd/dt → 0` produces divide-by-zero spikes. The `[0.05, 1.75]` validity window in config.toml filters both the zero spike and the `inf` outliers.

### `get_lh_power_threshold` → `lh_power_threshold`

L–H transition power threshold from the Martin 2008 scaling (Y. R. Martin et al 2008 *J. Phys.: Conf. Ser.* **123** 012033):

```
P_thr = 0.0488 · n_e^0.717 · B_t^0.803 · S^0.941   [MW]  →  ×1e6  [W]
```

with `n_e` in 10²⁰ m⁻³ (`get_densities`), `B_t` the baseline-subtracted absolute toroidal field (`get_btor`), and `S` the plasma surface area from EFIT (`\efit_aeqdsk:psurfa`, ~7.3 m² on C-Mod; the node has no units attribute in the tree but its values are m²). The `np.sign(x)·|x|^p` form guards against complex results from negative inputs (check the code).

image

Roughly log-normal distribution, peaked near ~1.2 MW with median ≈ 1.30e6 W and IQR ≈ [9.7e5, 1.6e6] W, which is where the measured C-Mod L–H threshold sits (~1–2 MW at 5.4 T). The histogram is clipped at 5 MW for display. A small number of unphysical outliers exist (min ≈ −7.0e6 W, max ≈ 1.5e7 W) driven by extreme density inputs: 3,680 of the 938,693 finite points (0.39%) fall outside the `[0.0, 1.0e7]` W validity range, 3,559 of them negative (0.38%) and 121 above 1e7 W.

## Changes after reviews

- **Naming**: `pow_thr_LH_Martin` → `lh_power_threshold` and `get_itpa_pow_thr` → `get_lh_power_threshold`. `run_columns` are lower-cased internally, so the mixed-case column name could not be requested at all.
- **`get_btor` → `bt`**, mirroring `D3DPhysicsMethods.get_btor`: `\btor` is now fetched and baseline-subtracted in one place and reused by `get_h98`, `get_lh_power_threshold` and `get_n_equal_1_amplitude` through the client-side cache, so it is read once per shot. `bt` keeps its name, units and config entry, it is just produced by `get_btor` instead of `get_n_equal_1_amplitude`. Checked on 100 random shots: `bt`, `h98` and `lh_power_threshold` bit-identical, `n_equal_1_mode`/`n_equal_1_normalized`/`n_equal_1_phase` agree to float32 round-off (max |Δ| 5.6e-8 T on the mode, ~1e-6 relative MAE, and 1e-13 relative when the same computation is redone in float64).
- **Surface area in the Martin scaling**: `S` now comes from `\efit_aeqdsk:psurfa` (plasma surface area, ~7.3 m²) Nothing else changed, the column is simply rescaled.
- **`h_alpha`**: lowercase MDSplus path, `try/except` removed (the framework already returns NaNs on failure), IMAS reference added in config.toml.
- **Validity literal** for `lh_power_threshold` written as `[0.0, 1.0e7]` (was `[0, 0.1e8]`, the same number in the style of the neighbouring entries).

## Warnings

- `h98` produces `inf` when `P_input ≈ dW_mhd/dt`; relies on user postprocessing.
- `h_alpha` can be negative and saturates at discrete values on some shots.
- Time base for `h_alpha` matters: use the native signal time base for ELM detection.
- `lh_power_threshold` is negative wherever the density input is negative (0.38% of points), since `B_t` and `S` enter as magnitudes; the validity range flags those.


disruption_py/machine/cmod/config.toml
```diff
@@ -81,6 +81,16 @@ imas = "/summary/global_quantities/greenwald_fraction/value"
 units = "dimensionless"
 validity = [0, 1.5]
 
+[cmod.physics.attributes.h98]
+description = "H98 confinement enhancement factor."
+units = "dimensionless"
+validity = [0.05, 1.75]
```
gtrevisan, line 88: imas?

disruption_py/machine/cmod/config.toml
```diff
@@ -81,6 +81,16 @@ imas = "/summary/global_quantities/greenwald_fraction/value"
...
+
+[cmod.physics.attributes.h_alpha]
+description = "H-alpha line emission intensity."
+units = "W/(m2*sr)"
+validity = [0.0, 161.920898]
```
gtrevisan, line 93: imas?

disruption_py/machine/cmod/config.toml
```diff
@@ -180,6 +190,11 @@ imas = "/bolometer/power_radiated_total"
 units = "W"
 validity = [0, inf]
 
+[cmod.physics.attributes.pow_thr_LH_Martin]
+description = "Martin 2008 L-H transition power threshold scaling."
+units = "W"
+validity = [0, 0.1e8]
```
gtrevisan, line 197: imas?

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        ----------
+        params : PhysicsMethodParams
+            The parameters containing the MDSplus connection, shot id and more.
+
+        Returns
+        -------
+        dict
+            A dictionary with the H-alpha signal (`h_alpha`). In IS brightness units [W/(m2*sr)].
+
+        Last major update by Enrique Zapata zapata_e@mit.edu
```
gtrevisan: you might want to reference this very PR -- take a look at the other docstrings for inspiration/formatting.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        """
+        output = {
```
gtrevisan: I don't think you need this: any unhandled exception will result in returning NaNs.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+            "h_alpha": [np.nan],
+        }
+        # Get signals from SPECTROSCOPY tree
+        try:
```
gtrevisan: you do not need the try block, the framework will catch and dump any exception as needed.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        dict
+            A dictionary with the H98 confinement enhancement factor (`h98`), which is dimensionless
+
+        References
+        ----------
+        Scaling from eq. 20, ITER Physics Basis Chapter 2
+        https://iopscience.iop.org/article/10.1088/0029-5515/39/12/302/pdf
+
+        Original Author: Andrew Maris (maris@mit.edu)
+        Last major update by Enrique Zapata (zapata_e@mit.edu)
```
gtrevisan: you might want to reference this very PR -- take a look at the other docstrings for inspiration/formatting.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+            A dictionary with the H-alpha signal (`h_alpha`). In IS brightness units [W/(m2*sr)].
+
+        Last major update by Enrique Zapata zapata_e@mit.edu
+        """
+        output = {
+            "h_alpha": [np.nan],
+        }
+        # Get signals from SPECTROSCOPY tree
+        try:
+            h_alpha, time_halpha = params.data_conn.get_data_with_dims(
```
gtrevisan: you should be able to invoke `get_data*` directly from `params`.
make sure you are based off the latest dev and remove `data_conn` from this call.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+                r"\SPECTROSCOPY::HA_2_BRIGHT", group="SPECTROSCOPY"
```
gtrevisan: although `group` is there and works, you should still use `tree` as usual.
also, I'm pretty sure we made all MDSplus paths lower case for consistency, LMK otherwise.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        """
+
+        # Get parameters for calculating confinement time
+        powers_df = CmodPhysicsMethods.get_power(params=params)
+        efit_df = CmodEfitMethods.get_efit_parameters(params=params)
+        density_df = CmodPhysicsMethods.get_densities(params=params)
+        ip_df = CmodPhysicsMethods.get_ip_parameters(params=params)
+
+        # Get btor
+        btor, t_mag = params.data_conn.get_data_with_dims(
```
gtrevisan: drop `data_conn`, use `tree` rather than `group`.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+            r"\btor", group="magnetics"
+        )  # tmag: [s]
```
gtrevisan: at this point mention units for both quantities, here, I guess T and s.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        # Toroidal power supply takes time to turn on, from ~ -1.8 and should be
+        # on by t=-1. So pick the time before that to calculate baseline
+        baseline_indices = np.where(t_mag <= -1.8)
```
gtrevisan, line 2184: don't you need a comma, here? please double check dimensions.
your slicing might work in the following line, but this might not be what you want -- a list of indices.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        dict
+            A dictionary with the power threshold for L-H transition
+            (`pow_thr_LH_Martin`), in Watts [W]
+
+        References
+        ----------
+        Scaling from Y R Martin et al 2008 J. Phys.: Conf. Ser. 123 012033
+        DOI 10.1088/1742-6596/123/1/012033
+
+        Last major update by Enrique Zapata
```
gtrevisan: same as above.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        """
+
+        density_df = CmodPhysicsMethods.get_densities(params=params)
+        areao_df = CmodPhysicsMethods.get_kappa_area(params=params)
+
+        # Get BT
+        btor, t_mag = params.data_conn.get_data_with_dims(
+            r"\btor", group="magnetics"
+        )  # tmag: [s]
```
gtrevisan: same as above.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        # Toroidal power supply takes time to turn on, from ~ -1.8 and should be
+        # on by t=-1. So pick the time before that to calculate baseline
+        baseline_indices = np.where(t_mag <= -1.8)
```
gtrevisan, line 2258: same as above.

gtrevisan (commented):
thanks, Enrique! it looks great.

I've left some minor comments after a quick look, I'll review again in the next few days.

disruption_py/machine/cmod/physics.py
yumouwei: I would remove the "Last major update..." line and add the "References" block like other physics methods.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        try:
+            h_alpha, time_halpha = params.data_conn.get_data_with_dims(
+                r"\SPECTROSCOPY::HA_2_BRIGHT", group="SPECTROSCOPY"
+            )  # [mW/(cm2*sr)], [s]
+            # Interpolate Halpha to params.times
+            h_alpha_interp = interp1(time_halpha, h_alpha, params.times)
+            output["h_alpha"] = 10 * h_alpha_interp  # [W/(m2*sr)]
+        except ValueError as e:
+            params.logger.warning("Failed to get H_alpha signal. Returning NaNs.")
+            params.logger.opt(exception=True).debug(e)
```
yumouwei: I don't think this `try-except` block would catch any `ValueError`. The only thing that could fail is R2140-2142 `params.data_conn.get_data_with_dims` (which can now be shortened to `params.get_data_with_dims`) which could be automatically caught by the framework.

Since the method will just return `output=[nan]` if R2140-2142 fails, we can safely  move R2140-2145 outside the `try-except` block, then remove R2135-2137 & R2146-2148.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+
+        Original Author: Andrew Maris (maris@mit.edu)
+        Last major update by Enrique Zapata (zapata_e@mit.edu)
+        """
+
+        # Get parameters for calculating confinement time
+        powers_df = CmodPhysicsMethods.get_power(params=params)
+        efit_df = CmodEfitMethods.get_efit_parameters(params=params)
+        density_df = CmodPhysicsMethods.get_densities(params=params)
+        ip_df = CmodPhysicsMethods.get_ip_parameters(params=params)
```
yumouwei: 1. All these methods return a dictionary, not a dataframe, so it'd be better to change the variable names to reflect their data type.
2. I'm always a little wary of calling other physics methods instead of fetching the raw signals directly from MDSplus, and then do some additional computation. The physics methods will return the signals already interpolated onto the requested timebase, so there could be a chance that this would introduce some errors when the original signals are sampled at a much lower frequency, or the timebase of the original signal is severely misaligned with the requested timebase. That being said I've done similar things before, and I don't think this is going to be a major problem here; it's just something to consider for future physics methods that involve using data from e.g. the Thomson system.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+
+        Returns
+        -------
+        dict
+            A dictionary with the H98 confinement enhancement factor (`h98`), which is dimensionless
+
+        References
+        ----------
+        Scaling from eq. 20, ITER Physics Basis Chapter 2
+        https://iopscience.iop.org/article/10.1088/0029-5515/39/12/302/pdf
```
yumouwei: I would make sure the format here is consistent with the References section in other physics methods' docstring. We've never referenced materials outside the disruption-py repo before; I think we can put this as a standalone bullet point and put the URL to a hyperlink, so something like this:

```
References
--------
- Scaling from eq. 20, [ITER Physics Basis Chapter 2](https://iopscience.iop.org/article/10.1088/0029-5515/39/12/302/pdf)
- pull requests: #[562](<link_to_this_pr>)
```

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        )
+        h98 = tau / tau_98
+        h98[h98 <= 0] = 0
+        output = {
+            "h98": h98,
+        }
+        return output
+
+    @staticmethod
+    @physics_method(columns=["pow_thr_LH_Martin"], tokamak=Tokamak.CMOD)
```
yumouwei: I'd use all lowercase for the signal name to align with the convention.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+
+        # Get BT
+        btor, t_mag = params.data_conn.get_data_with_dims(
+            r"\btor", group="magnetics"
+        )  # tmag: [s]
+        # Toroidal power supply takes time to turn on, from ~ -1.8 and should be
+        # on by t=-1. So pick the time before that to calculate baseline
+        baseline_indices = np.where(t_mag <= -1.8)
+        btor = btor - np.mean(btor[baseline_indices])
+        btor = np.abs(interp1(t_mag, btor, params.times))  # [T]
```
yumouwei, line 2260: We might want to distill the btor computation in `get_n_equal_1_amplitude`, `get_h98` and `get_itpa_pow_thr` into a single physics method similar to `D3DPhysicsMethods.get_btor`, or just call `get_n_equal_1_amplitude` to get the `btor` signal.

I do realize this suggestion is contradictory with my suggestion in R2180 so take this with a grain of salt.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        itpa_power_thr = (
+            0.0488
+            * (np.sign(n_e) * np.abs(n_e) ** 0.717)
+            * (np.sign(btor) * np.abs(btor) ** 0.803)
+            * (np.sign(areao) * np.abs(areao) ** 0.941)
+        )
+        output = {
+            "pow_thr_LH_Martin": 1e6 * itpa_power_thr,
+        }
+        return output
```
yumouwei: We can just do `return {"pow_thr_lh_martin": 1e6 * itpa_power_thr}`

yumouwei (changes requested):

disruption_py/machine/cmod/physics.py
yumouwei: I also just found out that `RetrievalSettings(run_columns=['pow_thr_LH_Martin']` doesn't call the method correctly unless I switch it to all lowercase

disruption_py/machine/cmod/config.toml
nbarbour13: [/summary/global_quantities/h_98](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/summary.html#summary-global_quantities-h_98)

disruption_py/machine/cmod/config.toml
nbarbour13: There may be a way to use the [`spectrometer_visible`](https://imas-data-dictionary.readthedocs.io/en/latest/generated/ids/spectrometer_visible.html) category for this.  I'm uncertain.

disruption_py/machine/cmod/config.toml
nbarbour13: I did not find a related IMAS entry.

disruption_py/machine/cmod/physics.py
nbarbour13: Understood and removed.

disruption_py/machine/cmod/physics.py
nbarbour13: Updated docstring with a link to this PR using the References block as recommended.

disruption_py/machine/cmod/physics.py
nbarbour13: Removed the try/except block.

disruption_py/machine/cmod/physics.py
nbarbour13: Updated the formatting of the docstring and included a link to this PR.

disruption_py/machine/cmod/physics.py
nbarbour13: Removed `data_conn`.

disruption_py/machine/cmod/physics.py
nbarbour13: Replaced `group` with `tree_name` and changed MDSplus path to lower case.

disruption_py/machine/cmod/physics.py
nbarbour13: Added unit T for `btor`.

disruption_py/machine/cmod/physics.py
nbarbour13: I believe this line is ok.
`np.where()` returns indices if no optional array arguments are provided.
I also checked the dimensions, but please confirm that it behaves as intended.

disruption_py/machine/cmod/physics.py
nbarbour13: Updated with a link to the article and the formatting changes requested.

disruption_py/machine/cmod/physics.py
nbarbour13: Resolved.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+            h_alpha, time_halpha = params.data_conn.get_data_with_dims(
+                r"\SPECTROSCOPY::HA_2_BRIGHT", group="SPECTROSCOPY"
+            )  # [mW/(cm2*sr)], [s]
+            # Interpolate Halpha to params.times
+            h_alpha_interp = interp1(time_halpha, h_alpha, params.times)
+            output["h_alpha"] = 10 * h_alpha_interp  # [W/(m2*sr)]
+        except ValueError as e:
+            params.logger.warning("Failed to get H_alpha signal. Returning NaNs.")
+            params.logger.opt(exception=True).debug(e)
+        return output
```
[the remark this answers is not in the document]
nbarbour13: Try/except block removed per Greg's suggestion.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        # on by t=-1. So pick the time before that to calculate baseline
+        baseline_indices = np.where(t_mag <= -1.8)
+        btor = btor - np.mean(btor[baseline_indices])
+        btor = np.abs(interp1(t_mag, btor, params.times))
+
+        # Get signals
+        ip = np.abs(ip_df.get("ip")) / 1.0e6  # [A] -> [MA]
+        n_e = density_df.get("n_e") / 1.0e19  # [m^-3] -> [10^19 m^-3]
+        p_input = powers_df.get("p_input") / 1.0e6  # [W] -> [MW]
+        dwmhd_dt = efit_df.get("dwmhd_dt") / 1.0e6  # [W] -> [MW]
```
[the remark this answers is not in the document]
nbarbour13: A significant portion of the data used to compute the IPB98(y,2) scaling were published in the second version of the H mode confinement database ITERHDB.2 (DB2).  An ITER H Mode Database Working Group paper detailing updates to this database was published in [Nuclear Fusion in 1994](https://iopscience.iop.org/article/10.1088/0029-5515/34/1/I10/pdf).  The units listed for line averaged density are 10^19 m^-3, and the paper lists two scaling laws for energy confinement time.
For reference, the IPB98(y,2) scaling can be found in Table 5 of [ITER Physics Expert Group on Confinement and Transport et al 1999 Nucl. Fusion 39 2175](https://iopscience.iop.org/article/10.1088/0029-5515/39/12/302/pdf).  This paper also lists 10^19 m^-3 as units of density.

I'll leave this thread open in case a reviewer wants to take a second look.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2107,167 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+            r"\btor", group="magnetics"
+        )  # tmag: [s]
+        # Toroidal power supply takes time to turn on, from ~ -1.8 and should be
+        # on by t=-1. So pick the time before that to calculate baseline
+        baseline_indices = np.where(t_mag <= -1.8)
+        btor = btor - np.mean(btor[baseline_indices])
+        btor = np.abs(interp1(t_mag, btor, params.times))  # [T]
+        n_e = density_df.get("n_e") / 1.0e20  # [m^-3] -> [10^20 m^-3]
+        areao = areao_df.get("kappa_area")  # [m^2]
+
```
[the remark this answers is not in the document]
nbarbour13: Fixed.

disruption_py/machine/cmod/physics.py
nbarbour13: Implemented these changes.

disruption_py/machine/cmod/physics.py
nbarbour13: Changed variable names from `*_df` to `*_dict`.
I retained the calls to the other physics methods, as you said it should not be a major problem here.

disruption_py/machine/cmod/physics.py
nbarbour13: Done.  Added a Markdown reference to satisfy the linting requirement of less than 100 characters per line.

disruption_py/machine/cmod/physics.py
nbarbour13: Changed name to lowercase in `config.toml` and here.

disruption_py/machine/cmod/physics.py
nbarbour13: No action was taken on this.

disruption_py/machine/cmod/physics.py
nbarbour13: Agreed.  Done.

disruption_py/machine/cmod/physics.py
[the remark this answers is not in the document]
yumouwei: @zapatace Can you make sure this is correct?

disruption_py/machine/cmod/physics.py
yumouwei: This is interesting. It looks like `arr[idx]` automatically unpacks the tuple, so `btor[baseline_indices]` automatically becomes `btor[baseline_indices[0]]`. I actually didn't know this, which makes me wonder why we chose to do `idx, = np.where(...)` in the first place...

yumouwei (approved):

disruption_py/machine/cmod/physics.py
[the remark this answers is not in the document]
zapatace: I remember wrestling with this. Thank you, @nbarbour13, for tracking down the references again! The papers state it pretty clearly. This H98 scaling is formulated with cental line averaged densities in [10¹⁹ m⁻³]. Here is some more evidence: [updated ITPA DB](https://www.doi.org/10.1088/1741-4326/abdb91). Copilot is hallucinating here. While it is true that in power threshold scaling the density is [10²⁰ m⁻³], this is not the case in H98.

disruption_py/machine/cmod/config.toml
zapatace: We can use 

```
[cmod.physics.attributes.h_alpha]
description = "H-alpha line emission intensity."
imas = "/spectrometer_visible/channel(i1)/grating_spectrometer/processed_line(i2)/radiance"
units = "W/(m^2*sr)"
validity = [0.0, 161.920898]
```

However, IMAS units are photons·m⁻²·s⁻¹·sr⁻¹ Do you want to use these units in the method (they are not SI)?

disruption_py/machine/cmod/config.toml
zapatace: Me neither. The only L–H-related IDSs are global_quantities/h_mode (confinement state label) and global_quantities/power_loss (actual P_sep), neither of which is the Martin-scaling threshold.

disruption_py/machine/cmod/physics.py
zapatace: I can make a helper function for the 3 methods. However, since I need to modify get_n_equal_1_amplitude can we leave this for a follow-up small PR? Up to you!

disruption_py/machine/cmod/physics.py
zapatace: You can see that behaviour is the same with this code snippet. 
```
import numpy as np
t_mag = np.linspace(-2.0, 0.0, 11)
btor  = np.arange(11, dtype=float)

baseline_indices = np.where(t_mag <= -1.8)   # tuple -> (array([0, 1]),)
idx, = np.where(t_mag <= -1.8)                # comma idiom -> array([0, 1])

np.array_equal(btor[baseline_indices], btor[idx])   # True
np.mean(btor[baseline_indices]) == np.mean(btor[idx])  # True (both 0.5)
```

disruption_py/machine/cmod/config.toml
gtrevisan: definitely use SI, IMAS is only for reference at this stage.
double check your units then solve the comment if resolved.

disruption_py/machine/cmod/physics.py
gtrevisan: I am aware, but you are calling that variable `idx` or `indices`, not `idx_tuple` or `indices_tuple`, so we can avoid confusion by using the comma. anybody handling that variable in the future won't have to extract indices from the tuple!

disruption_py/machine/cmod/physics.py
gtrevisan: same as above.

disruption_py/machine/cmod/physics.py
gtrevisan: maybe we should chat about this, then.

zapatace: I added a new common method get_btor.

disruption_py/machine/cmod/config.toml
```diff
@@ -107,6 +119,11 @@ imas = "/summary/boundary/elongation"
 units = "dimensionless"
 validity = [0.8, 2.0]
 
+[cmod.physics.attributes.lh_power_threshold]
+description = "Martin 2008 L-H transition power threshold scaling."
+units = "W"
+validity = [0.0, 0.1e8]
```
gtrevisan, line 125: was there a specific reason for writing `0.1e8` rather than simply `1.0e7`?

disruption_py/machine/cmod/physics.py
gtrevisan, line 935: I'd imagine here you'll explicitly call your new `get_btor` method.

disruption_py/machine/cmod/physics.py
```diff
@@ -902,7 +937,7 @@ def get_n_equal_1_amplitude(params: PhysicsMethodParams):
         )  # [T], [s]
         # Toroidal power supply takes time to turn on, from ~ -1.8 and should be
         # on by t=-1. So pick the time before that to calculate baseline
-        baseline_indices = np.where(t_mag <= -1.8)
+        (baseline_indices,) = np.where(t_mag <= -1.8)
```
gtrevisan, line 940: likewise, I'd imagine that by calling `get_btor` you'll get a `bt` already manipulated to account for this.

gtrevisan (changes requested):
AFAICS, you need to call your new `get_btor` method from `get_n_equal_1_amplitude` in order to leverage our client-side caching logic.

Note: I've found another example of `\btor` getting pulled from `get_te_profile_params_ece`, but that does not subtract a baseline and is used for a different purpose, so we might want to leave it as is and rely instead on server-side caching.

disruption_py/machine/cmod/config.toml
zapatace: Changed to validity = [0.0, 1.0e7]

disruption_py/machine/cmod/physics.py
zapatace: get_n_equal_1_amplitude no longer reads \btor: it takes bt = CmodPhysicsMethods.get_btor(params=params)["bt"] (baseline-subtracted, cached), so \btor is now read once per shot. 

Because get_btor returns the field already interpolated, the per-sensor order became baseline-subtract → interpolate → subtract pickup (previously pickup was subtracted before interpolating); that's exact for linear interpolation. Each BP13 sensor is now fetched with its own timebase (get_data_with_dims).

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2138,150 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        h98 = tau / tau_98
+        h98[h98 <= 0] = 0
+        output = {
+            "h98": h98,
+        }
+        return output
+
+    @staticmethod
+    @physics_method(columns=["lh_power_threshold"], tokamak=Tokamak.CMOD)
+    def get_lh_power_threshold(params: PhysicsMethodParams):
```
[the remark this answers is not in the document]
zapatace: I updated PR docs changing the naming, I also regenerated reference shot plots and distributions.

disruption_py/machine/cmod/physics.py
```diff
@@ -2106,3 +2138,150 @@ def _is_on_blacklist(shot_id: int) -> bool:
...
+        - pull requests: #[562](https://github.com/MIT-PSFC/disruption-py/pull/562)
+
+        [martin_reference]: https://iopscience.iop.org/article/10.1088/1742-6596/123/1/012033/pdf
+        """
+
+        density_dict = CmodPhysicsMethods.get_densities(params=params)
+        # Get area and interpolate to time base
+        area, t_aeqdsk = params.get_data_with_dims(
+            r"\efit_aeqdsk:areao/1e4", tree_name="_efit_tree"
+        )  # area: [m^2], t_aeqdsk: [s]
```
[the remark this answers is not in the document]
zapatace: Good catch! I'll use EFIT `\efit_aeqdsk:psurfa` from now on. I've replotted the reference shots and distributions and it all makes sense now. Hopefully, the third time will be the charm.

zapatace:
For the records. Knowing that methods involving Mirnovs are precision sensitive. I made the exercise of comparing outputs before and after changes in 77 shots when I changed get_btor.

As I expected, the data is not identical, but within a reasonable margin of numerical noise.

| column | unit | MAE | RMSE | max abs diff | mean abs (reference) | n shots finite |
| --- | --- | --- | --- | --- | --- | --- |
| n_equal_1_mode | T | 5.2423e-09 | 8.5843e-09 | 5.5879e-08 | 5.2640e-03 | 77 |
| n_equal_1_normalized | - | 1.0062e-09 | 1.6717e-09 | 1.5367e-08 | 1.0282e-03 | 77 |
| n_equal_1_phase | rad | 1.9311e-06 | 4.7289e-06 | 1.4707e-04 | 1.4764e+00 | 77 |
| bt | T | 0.0000e+00 | 0.0000e+00 | 0.0000e+00 | 5.2599e+00 | 100 |
| h98 | - | 0.0000e+00 | 0.0000e+00 | 0.0000e+00 | 2.6640e+06 | 98 |
| lh_power_threshold | W | 0.0000e+00 | 0.0000e+00 | 0.0000e+00 | 5.0766e+04 | 100 |


image
image

gtrevisan (approved):
LGTM! let's wait for Enrique to address the last comment from Copilot, then we can merge.

PR #564, opened 2026-06-04
Add C-Mod Thermal Quench Labeler

hwietfeldt:
# Overview
I created a C-Mod physics method `get_thermal_quench_time()` that labels the onset of the thermal quench (TQ) for a given shot (NaN for non-disruptive shots) using a vertical SXR array. The TQ time is returned as a column with identical values for each time slice. I used SXR instead of ECE because ECE is less reliable (especially for shots at low Bt) and because the SXR arrays can track the plasma core during a hot VDE. I used a vertical array because the horizontal array wasn't working for most 2005 shots. For 100 manually labeled flattop TQ onsets, this method labeled the onset within 1 ms of the manual label in 90% of cases. For the remaining 10%, the automated TQ label is usually still a more accurate disruption time metric compared to the current t_disrupt based on the maximum of |dIp/dt|. This method has also worked well in my UFO labeling workflow. **I have not tested this method during ramp-up or ramp-down**

# Steps of Method
1. Read Ip and SXR chords from MDSplus
2. Subtract background from each SXR chord
3. Discard bad chords dominated by noise by removing chords with low autocorrelation.
4. Filter the SXR chords using a non-causal Butterworth low pass filter.  
This helps to smooth transient, positive SXR spikes during the TQ, which I attribute to recombination radiation. There is a tradeoff when choosing the cutoff frequency between maintaining decent temporal resolution and removing the recombination spikes. I found a cutoff frequency of 1 kHz works best.
5. Select the SXR chord viewing the core by taking the maximum of SXR across all chords for each time slice.
6. Search for the onset of the CQ based on Ip and t_disrupt.  
This enables us to search for the TQ in a small time window prior to the CQ onset. We need a small search window to avoid labeling sawtooth crashes as the TQ while enabling us to label the TQ for shots with long CQs (see shots 1050311013, 1050802017).
7. Find the midpoint of the TQ by finding the minimum of dSXR/dt within a 5 ms window prior to the CQ onset.
8. Find the TQ onset within a 0.5 ms window prior to the TQ midpoint.  
Note that almost all TQs have a duration less than 1 ms.

# Case that works well

good_case

We correctly label the TQ onset, which is around 7 ms prior to $t_{disrupt}$. The $T_{e0}$ signal is junk because ECE was not working on this shot, which motivates the use of SXR for TQ labeling on C-Mod.

# Results from Testing

man_labels_test

I tested this method on 100 disruptive shots for which I manually labeled the TQ onset and end using ECE, SXR, H alpha, and other diagnostics. The plot below compared the error in the automated TQ onset vs the error we currently get when we use t_disrupt as the onset of the disruption compared to my manually labeled TQ onsets. From the y-values, we see that the automated TQ labeling routine labeled the TQ onset within 0.5 ms of the actual onset for most shots. As guaranteed by the algorithm, no automated TQ labels are later than t_disrupt. There are a several labels that are early or late by more than a  ms. In these cases, there may be some ambiguity in the TQ onset (multiple partial thermal collapses), or there were strong recombination spikes. The search windows and filtering cutoff frequency have tradeoffs, so I was not able to eliminate these outliers. Nevertheless, I believe that the TQ labels provide a much more accurate metric of the onset time of the disruption for many of our workflows compared to the current t_disrupt.

# Most early label
most_early_label

In this shot, the TQ time is automatically labeled around 0.511 s, which is at the onset of a sawtooth or small thermal collapse. I manually labeled the TQ onset around 0.514 s, which is the main thermal quench according to ECE (second panel) and the Ip spike (first panel). Again, I don't think that is 3 ms early label for the TQ is too bad. Using t_disrupt may be worse for some workflows as it is clearly after the disruption onset. If using t_disrupt, the user risks leaking disruption information into their disruption prediction model.

# Most late label
Screenshot 2026-06-04 at 3 21 13 PM

Here a very strong SXR recombination spike coupled with a long delay between the Ip spike and when Ip crashes causes a late label. The TQ onset is not much better than t_disrupt, but it's also not worse.

# Timing

I timed various steps of the TQ labeler. It can be slow for disruptive shots from 2012-2016. Note that for 2005 shots, the SXR was digitized at 5 kHz, not 250 kHz, so the algorithm is faster. Reading all 38 chords from MDSplus is a bottleneck. I tried using MDSplus's `getMany()` function to read the chords in one batch call; however, that was slower. For non-disruptive shots, the algorithm is fast because it immediately returns NaNs if t_disrupt is 0.

Timing averages for 2005 shots
Reading MDSplus: 0.20789 s
Calculating autocorrelation: 0.0264536 s
Butterworth Filter: 0.00573944 s
Finding TQ: 0.000120203 s
Total: 0.242485 s

Timing averages for 2012-2016 shots
Reading MDSplus: 3.87387 s
Calculating autocorrelation: 0.0920277 s
Butterworth Filter: 0.194109 s
Finding TQ: 0.00131993 s
Total: 4.16501 s


hwietfeldt:
@yumouwei @AlexSaperstein 

Once we are satisfied with the method and testing, I'll clean up the method and turn the draft PR into a normal PR.

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,172 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+
+        Returns
+        ----------
+        thermal_quench_time : array_like
+            time of thermal quench onset for the shot, identical values at each time-slice
+
+        References
+        ----------
+        - pull requests:
+        - issues:
```
yumouwei: @hwietfeldt can you fill in these references?

yumouwei (changes requested):
I refactored the lines for retrieving the SXR chords and doing the autocorrelation check. Can you double-check and make sure this method still behaves the same?

disruption_py/machine/cmod/physics.py
hwietfeldt: I reproduced the previous test results with your new commits (see the below plot). I'll add the references soon.
compare_tq_test_results_william_commits

hwietfeldt: Note the similar plot from the initial comment on the PR was from a slightly outdated version, so the results of the 'Old' are a bit better than the test results of the plot in the initial comment. 'Old' refers to the test results of the latest version prior to William's commits, and 'New' is using the version after William's commits. They agree perfectly, so the method behaves the same before and after William's refactoring.

AlexSaperstein (approved):
Things look good to me. I checked the TQ labels against the manual labels i did for the 100 flattop shots from ramping analysis and the labels for the most part look much better than the automated t_disrupt (CQ time) labels
Image
The caveat is that within this dataset there are more examples where the TQ is labeled more than 1 ms beyond the manual label (31 out of the 100 cases). However, after manually inspecting several of the outlier cases, it seems that many of their manual labels were maybe just a little too early, with the automatic TQ label appearing more appropriate.
Image
As a result, it looks to me like these labels are pretty good!

disruption_py/machine/cmod/physics.py
yumouwei: @hwietfeldt can you put this plot and the comments in the main thread? This comment thread will be collapsed once we hit "resolve conversation"

hwietfeldt:
I reproduced the previous test results with William's new commits. Note the similar plot from the initial comment on the PR was from a slightly outdated version, so the results of the 'Old' are a bit better than the test results of the plot in the initial comment. 'Old' refers to the test results of the latest version prior to William's commits, and 'New' is using the version after William's commits. They agree perfectly, so the method behaves the same before and after William's refactoring.
compare_tq_test_results_william_commits

hwietfeldt:
I added the references and a sentence describing the accuracy of the labeler to the docstring. I also merged the latest few commits from dev. I reran my testing scripts, reproducing the above plot, this time with 'Old' being the results after William's commits and 'New' being after merging the latest commits from dev, so the labeler works as before.

yumouwei (approved):

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
 
         return {"v_surf": v_surf}
 
+    @staticmethod
+    @physics_method(columns=["thermal_quench_time"], tokamak=Tokamak.CMOD)
+    def get_thermal_quench_onset_time(params: PhysicsMethodParams):
```
gtrevisan: we should drop `onset` from this method's name in order to mimic what we do for `get_current_quench_time`.
```suggestion
    def get_thermal_quench_time(params: PhysicsMethodParams):
```

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+            time of thermal quench onset for the shot, identical values at each time-slice
+
+        References
+        ----------
+        - pull requests: #[564](https://github.com/MIT-PSFC/disruption-py/pull/564)
+        - issues: #[542](https://github.com/MIT-PSFC/disruption-py/issues/542)
+
+        """
+        # Skip labeling the thermal quench time if the shot is non-disruptive
+        if params.disruption_time is None:
```
gtrevisan, line 2133: this is using the disruption time as read from the `disruptions` SQL table.
is this what we want?
I'd imagine we want to rely on our python-controlled current quench time, now.

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+        t_sxr = None
+        for i in range(n_chords):
+            try:
+                chord, t_chord = params.get_data_with_dims(
+                    f"{array_path}:chord_{i+1:02}",
+                    tree_name="xtomo",
+                )
+            except mdsExceptions.MdsException:
+                params.logger.debug(
+                    "get_thermal_quench_onset_time: "
```
gtrevisan:
```suggestion
                    "get_thermal_quench_time: "
```

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+                    f"Failed to get SXR {array_path} chord {i+1} data."
+                )
+                continue
+            # Subtract constant background
+            chord = chord - np.mean(chord[t_chord < 0.0])
+            valid = (t_chord > 0) & (t_chord < params.disruption_time + 0.05)
```
gtrevisan, line 2159: same logic here, to be discussed.
ideally you'd like to see the different between the DB disruption time and our python CQ time for a handful of your shots.

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+            if t_sxr is None:
+                t_sxr = t_chord[valid]
+            # Occasionally the time bases of a chord are of a different length
+            # Usually one timebase is just cut off early after shot is over
+            if np.sum(valid) == len(t_sxr):
+                chords[i] = chord[valid]
+
+        if t_sxr is None:
```
gtrevisan, line 2167: I think here you can raise maybe a `NanDataError` with an explicit error message.
the result will be the same, that is, NaNs all around.

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+        sample_time = t_sxr[1] - t_sxr[0]
+        sample_freq = 1 / sample_time
+
+        # Remove bad chords by checking each chord's autocorrelation.
+        # Bad chords often have significant white noise, meaning low autocorrelation (< 10 ms)
+        # Good chords should have an autocorrelation of 100s of ms
+        # See shot 1050311013 as an example with some bad chords
+        noise_autocorr_cutoff = 0.01  # [s]
+        # Use 300 ms prior to current quench for speed-up during autocorr O(N^2)
+        idx_start = np.argmin(np.abs(t_sxr - (params.disruption_time - 0.3)))
```
gtrevisan, line 2182: and again -- you are using the SQL-based `disruption_time` extensively, yet its calculation is not within our control. we might want to discuss this.

disruption_py/machine/cmod/physics.py
```diff
@@ -8,6 +8,7 @@
 
 import numpy as np
 import scipy.constants as const
+from scipy.signal import butter, filtfilt, resample_poly
```
gtrevisan: I think it's better to simply `import scipy` and then use the full qualifiers for each package.
it makes it easier to track scipy usage and possible deprecations in the future.

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+            index_no_lag = np.argmax(autocorr)
+            crosses_zero = autocorr[index_no_lag:] < 0
+            # See shot 1120223007 for example of why this fallback is necessary
+            index_decay = (
+                np.argmax(crosses_zero) if np.any(crosses_zero) else len(crosses_zero)
+            )
+            autocorr_decay_time = index_decay / sample_freq_5khz
+            if autocorr_decay_time < noise_autocorr_cutoff:
+                params.logger.debug(
+                    "get_thermal_quench_onset_time: "
```
gtrevisan:
```suggestion
                    "get_thermal_quench_time: "
```

disruption_py/machine/cmod/physics.py
```diff
@@ -2094,6 +2095,175 @@ def get_surface_voltage(params: PhysicsMethodParams):
...
+        # Search for TQ midpoint as min(dSXR/dt) in window of 5 ms prior to current quench onset
+        wndw_before_cq = 0.005  # [s]
+        idx_start = np.argmin(np.abs(t_sxr - (cq_onset_time - wndw_before_cq)))
+        idx_end = np.argmin(np.abs(t_sxr - cq_onset_time))
+        if idx_start == len(t_sxr) - 1:
+            params.logger.debug(
+                "get_thermal_quench_onset_time: No SXR data at time of CQ."
+                f"params.disruption_time = {params.disruption_time:.3f}."
+            )
+            return {"thermal_quench_time": [np.nan]}
```
gtrevisan: please use the most appropriate custom exception here, too.

gtrevisan (commented):
I think we only need to discuss about the SQL-based disruption time vs our python-based current quench time, and then we can merge!

hwietfeldt:
I changed the scipy import to just one `import scipy` with full qualifiers for each package at each function call. In addition to changes in `get_thermal_quench_time()`, there was also a change in one line of `_get_te_profile_params_ece()`, which used `scipy.constants`. I performed a diff compare on DisruptionPy outputs for both functions on 1000 C-Mod shots before and after the changes to the scipy import. **Results were identical**

gtrevisan (commented):
after discussions with Henry, this is our to-do list:
- move from old-school `disruption_time` to our new `current_quench_time`
- evaluate cyclic import
- evaluate timing

hwietfeldt:
# Switch from SQL disruption time to Generic CQ time

I tested the differences in thermal quench time labels using the generic `current_quench_time` vs the SQL `disruption_time` on all C-Mod shots that reached flattop from 2012-2016 (~8000 shots). Below are plots similar to #545:

## More shots with thermal quench
cq_time_vs_t_disrupt_conf_matrix

Using the generic CQ method instead of the SQL `disruption_time`, more shots are labeled to have a TQ. This is because the generic CQ method labels more shots to have CQs than the routines used to build the SQL table containing `disruption_time` (see #545). A shot should be labeled to have a TQ if and only if it's labeled to have a CQ, so this behavior is desired.

## Differences in TQ time values are basically unchanged

cq_time_vs_t_disrupt_hist

The above plot is a histogram in the differences in the output `thermal_quench_time` when using the generic `current_quench_time` vs the SQL `disruption_time` for all 2840 shots in which both choices result in a labeled TQ. The differences are $\ll 1 \mu$s, which is much less than the error based on comparison to manual labels (see above posts), so these differences are negligible. This finding is expected because according to #545, differences in labeled CQ times are fractions of a ms, and the TQ time labeler should not be sensitive to differences in the CQ of < 1 ms.

hwietfeldt:
# Speed-up for Thermal Quench Labeling

The computation of the thermal quench time is limited by the read-time for the soft X-ray (SXR) chords.
The main issue is that for shots in the years 2012-2016, the SXR chords were digitized to 250 kHz,
resulting in very long arrays.

## Summary of speed-ups

**Prior to the speed-ups, the read time per disrupted shot (2012-2016) was 3.7 s on the MFE workstations**.
Note that for shots without a current quench (about half of shots), `get_thermal_quench_time` immediately returns NaN.

- Read the time base of only one chord
- Only read 21 chords instead of 38 (excluding the outer 17 views did not result in meaningful changes)
- Subtract SXR chord background within a TDI expression
- Read only snippets of SXR chords (only minor speed-up of 0.06 s / shot)
- vectorized autocorrelation calculation (minor speed-up; not relevant for reading signals)

**These improvements result in a 0.7 s read time per disrupted shot (2012-2016) on the MFE workstations.**
The first three bullets each significantly contribute to the speed-up.
The output `thermal_quench_time` is not meaningully different with these speed-ups for 50 shots tested.

## TDI Expression for reading chord data

The data for each chord is now read with a TDI expression that grabs only a snippet of the data and subtracts the background signal. This was found to be the fastest way to get background-corrected chord data.

```
sig = f"{array_path}:CHORD_{idx_first_chord+i+1:02}"
tdi_expr = f"""
    _s = data({sig});
    _y = _s[{j_chord_start} : {j_chord_end-1}]
        - mean(_s[{j_bgrnd_start} : {j_t0-1}]);
    _y
    """
chord = params.get_data(tdi_expr, tree_name="xtomo")

```

## Changes to output thermal quenches
For about 1400 flattop disruptions in the C-Mod disruption warning database, I calculated the differences in the output `thermal_quench_time` compared to those prior to the speed-up related commits. The vast majority of shots had no changes. There were five cases of shots in which the magnitude of the differences was over 1 ms. On these shots, the SNR of the SXR chords was low (so the algorithm struggles anyways), and minor diffs in the autocorrelation calculation on smaller snippets propagated into the output. **Importantly, there were no shots for which the difference was larger than the error range based on manual testing (see plot in previous comments).**

tq_changes

## Attempted speed-ups that did not work

### Resampling in TDI expression

I attempted to resample the signals with the following TDI expression:

```
tdi_expr = (
    f"data(resample(make_signal("
    f"data({sig})[{i_chord_start}:{i_chord_end-1}], *, "
    f"dim_of({sig})[{i_chord_start}:{i_chord_end-1}]), {t_wndw_start}, {t_wndw_end}, {0.001}))"
)
chord = params.get_data(tdi_expr, tree_name="xtomo")
```

This resulted in 1.25 s read time per disruptive shot compared to 0.8 s / shot in the final implementation.
The likely explanation is that the resampling function still needs to decompress the entire SXR signal before resampling and addtionally needs to read the time base.

### Batch call of SXR chords with fall-back

I attempted to get all chords in one TDI expression containing an array of `data()` calls; however, this did not speed-up the reads.

### Removing background subtraction

Background subtraction is necessary when filtering out SXR chords dominated by white noise (see C-Mod shot 1050311013).

gtrevisan:
I should have fixed the cyclic import with this yet-to-be-merged PR:
- #599

and a quick follow-up commit in this branch:
- https://github.com/MIT-PSFC/disruption-py/commit/96cf23a41cfffbdd41659a727b8bf83e91c34d9b

@hwietfeldt please take a look at this again, it should all work as expected.

note: `pylint` was failing on `test_config.py` and required disabling the `cyclic-import` code, too.

as a final step, I'll invoke Copilot, just in case...

hwietfeldt:
I addressed Co-pilot's requested changes:
- Typo in docstring
- Typo in logging string
- Mask out bad chords instead of setting values to 0

For the third change, I took a simpler, more readable approach than Co-pilot's suggestion by using a Boolean array `valid_chords` to mask out any chords without data or deemed noisy based on the chord's autocorrelation.

I tested these changes on the 1400 flattop disruptions in the disruption warning database on C-Mod. 
- No changes from Greg's commits
- The valid_chord mask caused changes to 3/1400 labeled thermal quench times. All three differences were < 0.2 ms, which is within the error of the labeler. These were cases in which a bad chord was used despite having 0 value (likely because the background subtraction subtracted slightly too much)

**In my opinion the branch is ready to merge.**

gtrevisan:
I ran this method again:
- `cmod_ufo` shotlist:
  - with a 1-cpu workflow:
     `Completed workflow: retrieved 122/122 shots (100.00%) in 1m 43s (0.847 s/shot)`
  - with a 8-cpu workflow:
     `Completed workflow: retrieved 122/122 shots (100.00%) in 17s (0.140 s/shot)`
- `disruption_warning` shotlist:
  - with a 16-cpu workflow:
    `Completed workflow: retrieved 10,429/10,435 shots (99.94%) in 4m 7s (0.024 s/shot)`

no errors were raised by the `disruption-errors` script.

timings are okay, so we can go ahead and merge...
... after a final Copilot review, since it always catches something new!

gtrevisan (approved):
alright! many thanks @hwietfeldt for pushing this through. 🚀

Issue #445, opened 2025-04-17
Evaluate proper design for configurations and settings


gtrevisan:
at present, there is some confusion between:
- configuration-based settings (eg generic config, or machine-specific test shots and columns), and
- code-based settings (eg default function arguments).

we should probably design this a bit better with maximum flexibility in mind.
I'm leaning towards making everything config-based, including, eg:
- log setting defaults,
- output setting defaults,
- nickname setting defaults,

most stuff could then be trivially overridden at runtime by [dynaconf's env vars logic](https://www.dynaconf.com/envvars/).

there might still be some complications, so a proper functional analysis is necessary.

samc24:
The prototype for this is up as #569: the `LogSettings` defaults (`file_level` plus the three console thresholds) now resolve from a `[default.log]` section in `config.toml`, with the full Dynaconf layering for free (explicit argument > `DISPY_LOG__*` env var > `user.toml` > repo `config.toml`). Details and the deliberate exclusions (`console_level`, `file_path`) are in the PR description.

For the functional analysis half of this issue, I surveyed the remaining hardcoded values. They fall into three tiers.

**A. Trivial scalars, same treatment as #569, no design questions.** The `LogSettings` tunables (done in #569), and the `RetrievalSettings` scalar defaults (`efit_nickname_setting="disruption"`, `time_setting="disruption_warning"`, `only_requested_columns=False`). Pure preference defaults with obvious config keys.

**B. Small design call needed first.** `LogSettings.file_path` (a computed template, so the config value would need placeholder semantics rather than a plain string), and the physics constants currently living as bare literals: the EAST 17 ms PCRL01 timebase shift (repeated in `time_setting.py`, `physics.py`, and `util.py`), the 100 ms bona fide plasma duration threshold, the HBT-EP 10 us timebase step. These are physics-meaningful numbers, so the question is whether they belong in user-facing config at all, or in a machine constants layer that is config-adjacent but not user-tunable.

**C. Architectural, should ride the settings revamp.** `OutputSetting` path logic, and the `NicknameSetting`/`TimeSetting` string-to-object registries. Config-izing these now would fight code that is actively changing shape, so I would fold them into the settings revamp rather than migrate them standalone.

[...]

PR #569, opened 2026-07-10
Resolve LogSettings defaults from configuration

samc24:
Prototype for #445: move hardcoded tunables into the layered configuration, using `LogSettings` as the proof of pattern.

- `file_level`, `console_level` and the console thresholds (`warning_threshold`, `success_threshold`, `info_threshold`) default to `None` and are resolved in `__post_init__` from the `[default.log]` section of `config.toml`. An unset `console_level` keeps the existing meaning: derive the level from the shot count.
- Precedence: explicit argument, then `DISPY_LOG__*` environment variable, then `~/.config/disruption-py/user.toml`, then the repo `config.toml`. The env and user-file layers come for free from the existing Dynaconf setup, e.g. `DISPY_LOG__CONSOLE_LEVEL=INFO`.
- The effective log settings are written back into the configuration inside `get_shots_data` before the per-run `config.json` dump, which now runs after the retrieval and output settings are resolved. The dump therefore reflects the runtime values for both CLI and API callers.
- The CLI `-l` default comes from the same config lookup instead of a hardcoded VERBOSE, so command-line runs get the dynamic level like the API. This is the one behaviour change.
- The file log level is floored at the console level: `-l TRACE` lowers both sinks, while a dynamic WARNING console on a large run leaves the file at DEBUG.
- `file_path` stays out of the config since it is a computed per-run default that needs its own design call. Log settings remain machine-independent by construction: logging is set up before the tokamak is resolved, so the `default` environment is the only possible source.
- Explicitly passed values always win, and the config defaults match the previous hardcoded values. Workers under spawn are unaffected since the settings are resolved in the main process before pickling.
- Tests in `tests/test_config.py` cover precedence, the config round-trip and write-back, and the file-level floor.

Not in this PR: the remaining CLI defaults (efit tree, time base, output, processes). Moving those means the dataclass defaults read config too and the resolved settings objects need a JSON form for the dump, which overlaps the settings revamp.

### DIFF
--- a/disruption_py/config.toml
+++ b/disruption_py/config.toml
@@ -1,3 +1,9 @@
+[default.log]
+file_level = "DEBUG"
+info_threshold = 50
+success_threshold = 500
+warning_threshold = 1000
+
 [default.tests]
 match_fraction = 0.95 # Fraction of signals that must match between MDSplus and SQL
 val_tolerance = 0.01 # Tolerance for comparing values between MDSplus and SQL
--- a/disruption_py/settings/log_settings.py
+++ b/disruption_py/settings/log_settings.py
@@ -15,6 +15,7 @@ from typing import Union
 from loguru import logger
 from tqdm.auto import tqdm
 
+from disruption_py.config import config
 from disruption_py.core.utils.misc import get_metadata, get_temporary_folder
 
 # Register logger state at module import time so it is available in every
@@ -48,13 +49,20 @@ class LogSettings:
     """
     Settings for configuring logging.
 
+    Defaults for `file_level` and the console thresholds are read from the
+    `[log]` section of the layered configuration (repo `config.toml`, then
+    `~/.config/disruption-py/user.toml`, then `DISPY_LOG__*` environment
+    variables, e.g. `DISPY_LOG__WARNING_THRESHOLD=2000`). Values passed
+    explicitly always take precedence over the configuration.
+
     Attributes
     ----------
     file_path : str, optional
         Path to the log file. If None, no log file will be created.
         By default, a log file will be created in a temporary folder.
-    file_level : str
-        Logging level for the log file (default is "DEBUG").
+    file_level : str, optional
+        Logging level for the log file. Default is None, so the level is read
+        from the configuration (`log.file_level`, "DEBUG" out of the box).
         Possible values are:
         "TRACE", "DEBUG", "VERBOSE" (custom), "INFO", "SUCCESS", "WARNING", "ERROR", "CRITICAL".
         See: https://loguru.readthedocs.io/en/stable/api/logger.html#levels
@@ -64,28 +72,45 @@ class LogSettings:
         Possible values are:
         "TRACE", "DEBUG", "VERBOSE" (custom), "INFO", "SUCCESS", "WARNING", "ERROR", "CRITICAL".
         See: https://loguru.readthedocs.io/en/stable/api/logger.html#levels
-    warning_threshold : int
+    warning_threshold : int, optional
         If number of shots is greater than this threshold, the console log level will
-        be "WARNING". Default is 1000.
-    success_threshold : int
+        be "WARNING". Default is None, so the value is read from the configuration
+        (`log.warning_threshold`, 1000 out of the box).
+    success_threshold : int, optional
         If number of shots is greater than this threshold and less than the warning_threshold,
-        the console log level will be "SUCCESS". Default is 500.
-    info_threshold : int
+        the console log level will be "SUCCESS". Default is None, so the value is read
+        from the configuration (`log.success_threshold`, 500 out of the box).
+    info_threshold : int, optional
         If number of shots is greater than this threshold and less than the success_threshold,
-        the console log level will be "INFO". Default is 50.
+        the console log level will be "INFO". Default is None, so the value is read
+        from the configuration (`log.info_threshold`, 50 out of the box).
     _logging_has_been_setup : bool
         Internal flag to prevent multiple setups (default is False).
     """
 
     file_path: str = os.path.join(get_temporary_folder(), "output.log")
-    file_level: str = "DEBUG"
+    file_level: str = None
     console_level: str = None
 
-    warning_threshold: int = 1000
-    success_threshold: int = 500
-    info_threshold: int = 50
+    warning_threshold: int = None
+    success_threshold: int = None
+    info_threshold: int = None
 
     _logging_has_been_setup: bool = False
+
+    def __post_init__(self):
+        # Resolve unset tunables from the layered config (repo config.toml,
+        # user.toml, DISPY_LOG__* environment variables). Explicit arguments
+        # always win.
+        log_config = config().log
+        if self.file_level is None:
+            self.file_level = log_config.file_level
+        if self.warning_threshold is None:
+            self.warning_threshold = log_config.warning_threshold
+        if self.success_threshold is None:
+            self.success_threshold = log_config.success_threshold
+        if self.info_threshold is None:
+            self.info_threshold = log_config.info_threshold
 
     def reset_handlers(self, num_shots: int = None):
         """
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -4,7 +4,8 @@
 
 import os
 
-from disruption_py.config import config
+from disruption_py.config import config, configs
+from disruption_py.settings import LogSettings
 
 
 def change_directory(test, tmpdir="/tmp"):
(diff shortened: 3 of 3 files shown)


gtrevisan (commented):
this works well in moving some of the previously-hardcoded values and logic into the settings.
(the following applies to log levels, too, even though it certainly does not matter much.)
because we chose log levels as a prototype for the other settings, we still have to refine:
- the behaviour of the [cli argparser hooks](https://github.com/MIT-PSFC/disruption-py/blob/166395a6eb5f869d9d7575cb251637f4ee30b869/disruption_py/workflow.py#L299), which do have a different set of defaults,
- the update of any cli setting into the config, so that our [config dump](https://github.com/MIT-PSFC/disruption-py/blob/166395a6eb5f869d9d7575cb251637f4ee30b869/disruption_py/workflow.py#L142) reflected the actual config at runtime.

we might want to also have all of our internal knobs as configs, eg including console level (default to unset because it's set dynamically, but does get stored if the user specifies it).

(ultimately, we need to have the config as the single source of truth for any configuration.)

gtrevisan:
(another quick feature for when you have some time:)

our file log level is always `DEBUG` -- can we make it so that it's always *at least* `DEBUG`?
in other words, if a user specifies `TRACE`:
- as is: console log level is lowered to `TRACE`, but file log level stays at `DEBUG`,
- to be: both console and file log levels are lowered to `TRACE`.

samc24:
hey @gtrevisan , i pushed a follow-up covering the review plus the file-level ask. PR desc updated to match.

- `console_level` is now a `[log]` config knob. Unset means dynamic as before, and a user.toml or `DISPY_LOG__CONSOLE_LEVEL` pins it.
- The CLI `-l` default reads that config value instead of a hardcoded VERBOSE, so CLI and API runs get the same dynamic level. That is the one behaviour change.
- The effective log settings are written back into the config before the dump, and the dump now runs after the retrieval and output settings resolve. I put the write-back inside `get_shots_data` rather than in the parser so it also covers API callers passing a `LogSettings`; a `config.update(vars(args))` in `cli()` would only catch the CLI path. Only the log settings are written back here, see below.
- File level is floored at the console level: `-l TRACE` lowers both sinks, a dynamic WARNING console on a large run leaves the file at DEBUG.

Left for a separate PR: the other CLI defaults (efit tree, time base, output, processes) and their write-back. They already match the API defaults, and doing them properly means the dataclass defaults read config too and the resolved settings objects need a JSON form for the dump, which overlaps the settings revamp.

Thoughts?

gtrevisan:
thanks! I will try and test it out in the next few days.

were you also able to dissect the `tqdm` vs `loguru` issues?
I believe our logging calls behave well in the physics methods due to our custom `tqdm.write` monkey patch in the `reset_handlers` method, but they don't behave as well throughout the whole framework.
as a consequence, our `tqdm` progress bar pollutes the console output by leaving its partial traces:
```
$ uv run disruption-py -l info 1150805012 1150805014

09:29:58.959 [ INFO  ] Starting workflow: 2 shots / 1 process
  0%|                                                                                                                                  | 0/2 [00:00<?, ?it/s]09:29:59.001 [ INFO  ] Database driver fallback: 'ODBC' -> 'ODBC Driver 18 for SQL Server'
09:30:00.548 [SUCCESS] #1150805012 | Success! Most parameters have data: 61/64 (95.31%) in 1s
 50%|█████████████████████████████████████████████████████████████                                                             | 1/2 [00:01<00:01,  1.59s/it]09:30:01.646 [SUCCESS] #1150805014 | Success! Most parameters have data: 60/64 (93.75%) in 1s
09:30:01.677 [ INFO  ] Completed workflow: retrieved 2/2 shots (100.00%) in 2s (1.359 s/shot)                                                              
```
(in other words, some `loguru` calls printed newlines that prevented `tqdm` from cleanly carrying its progress bar over to the next line.)

samc24:
on tqdm: had a first look. i think the patch itself is fine, i think the catch is it only helps in the process that owns the bar. `tqdm.write` clears and redraws the bars it knows about, and in a worker there aren't any (the bar lives in the main process, `tqdm._instances` is empty in the worker), so from a worker `tqdm.write` is just a print that lands wherever the cursor is. in your trace every line glued to the bar is a worker message (the worker-side DB fallback, the per-shot SUCCESS lines), while "Completed workflow" from the main process comes out clean. so i'd say the split is main vs worker rather than physics methods vs framework, the physics-method logs come from workers too. 
that's #588, and i'd keep it out of this PR: it needs the workers to hand console output to the main process, which is a different change from the config plumbing here. separate PR off dev once this one is in.

for testing this one, the quick version on your shots:

    uv run disruption-py -l trace 1150805012                              # TRACE on console AND in output.log
    DISPY_LOG__CONSOLE_LEVEL=WARNING uv run disruption-py 1150805012      # quiet console, output.log still DEBUG
    uv run disruption-py 1150805012                                       # dynamic as before

then `LOG` in the run's config.json should show `"TRACE"`, `"WARNING"` and `null` respectively.

no rush, just let me know if you'd rather i change anything before you get to it. @gtrevisan

disruption_py/settings/log_settings.py
```diff
@@ -48,45 +49,101 @@ class LogSettings:
     """
     Settings for configuring logging.
 
+    Defaults for `file_level`, `console_level` and the console thresholds are
+    read from the `[log]` section of the layered configuration (repo
+    `config.toml`, then `~/.config/disruption-py/user.toml`, then `DISPY_LOG__*`
```
gtrevisan, line 54: while this is the actual reading order, it's the *reversed* order of precedence, so you might want to tweak wording on your comment just to be clear for careless readers.

disruption_py/settings/log_settings.py
```diff
@@ -136,7 +193,7 @@ def reset_handlers(self, num_shots: int = None):
             is_main = multiprocessing.current_process().name == "MainProcess"
             logger.add(
                 self.file_path,
-                level=self.file_level,
+                level=resolve_file_level(self.file_level, console_level),
```
gtrevisan, line 196: this is my only main comment -- this resolution happens on-the-fly in-memory and does not get mapped/persisted in the configuration, so eg any configuration dump will probably show file log level as being still DEBUG.
you should find the best-practice pattern and update that config value somewhere along the route.

disruption_py/config.toml
```diff
@@ -1,3 +1,10 @@
+[default.log]
+# console_level = "VERBOSE" # unset: chosen dynamically from the number of shots
```
gtrevisan, line 2: the dynamic levels were useful when we had fewer controls, but we could also remove them altogether now that we are revamping it. your call!
if you were a user, would you like to have these thresholds in?

gtrevisan (commented):
(now that I understand the problem, I agree with tackling #588 separately.)

only one comment, the rest being suggestions for simplification -- eg dropping dynamic console levels.

disruption_py/settings/log_settings.py
samc24: reworded, the docstring now lists the precedence order (argument, env, user.toml, repo config) instead of the reading order. same fix on the comment in `__post_init__`.

disruption_py/settings/log_settings.py
samc24: agreed, the min was only feeding `logger.add`, so the object said DEBUG while the sink was at TRACE. the floor now happens once in `__post_init__`, so `file_level` on the object is the effective value, the handlers in the main process and in the workers read it, and `to_config()` writes it into the dump. that is the same shape as disruption-efit, resolve the value then `config.update` it, only here the resolved object is the carrier so API callers get it too. `resolve_file_level` returns the level itself instead of its number so the dump keeps the name. `-l trace` gives `"file_level": "TRACE"` and `"console_level": "TRACE"` in config.json, `-l warning` leaves the file at `"DEBUG"`.

disruption_py/config.toml
samc24: dropped them. tbh from a user perspective IMO I want a sane default for a couple of shots, for a big run I set the level anyway, and I would never tune three thresholds. the value was in the default, not in the knobs, and now the default is one config key that user.toml or `DISPY_LOG__CONSOLE_LEVEL` can override.

fixed default is INFO rather than the old VERBOSE: the VERBOSE lines are the per-method "no data found" notes, useful on one shot but noise on ten...
the one thing to be aware of is that the per-shot summary line ("Success! all parameters have data") is at SUCCESS, which sits above INFO, so a 60k-shot run that sets nothing now prints one line per shot where the dynamic WARNING used to hide them. `-l warning` or `console_level = "WARNING"` in user.toml gives the old behaviour for big runs. this also deviates from the previous "default to unset" idea, `console_level` is now always set, which is what makes the dump honest without waiting for the shot count.

disruption_py/workflow.py
```diff
@@ -132,6 +132,17 @@ def get_shots_data(
...
+    # Clean-up parameters
+    if retrieval_settings is None:
+        retrieval_settings = RetrievalSettings()
+
+    retrieval_settings.resolve()
+    output_setting = resolve_output_setting(output_setting)
+
+    # write the effective settings back into the configuration, so that the
+    # dump below reflects the actual runtime values rather than the file defaults
+    config(tokamak).update({"log": log_settings.to_config()})
```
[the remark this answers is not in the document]
samc24: resolved by dropping the dynamic level: `console_level` is fixed at construction and the floor is applied there too, so the write-back before the dump carries both effective values.

tests/test_config.py
```diff
@@ -36,3 +38,86 @@ def test_access_tokamak_settings(tokamak):
...
+    monkeypatch.setenv("DISPY_LOG__CONSOLE_LEVEL", "WARNING")
+    # drop the cached config so the environment variables are re-read
+    configs.pop("default", None)
+    try:
+        settings = LogSettings()
+        assert settings.warning_threshold == 2000
+        assert settings.file_level == "INFO"
+        assert settings.console_level == "WARNING"
+        # untouched fields still come from the config file
+        assert settings.success_threshold == 500
```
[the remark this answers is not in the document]
samc24: the thresholds are gone. the remaining untouched-field check compares against the layered config.

samc24:
@gtrevisan 
i just pushed the follow-up attempt: dynamic thresholds dropped, `console_level = "INFO"` in config.toml, file-level floor applied in `__post_init__` so the object, the handlers and the dump agree. net about 50 lines fewer. stale threshold keys in an existing user.toml are ignored (they just show up in the dump as unused keys). same three commands as before to test, config.json `LOG` should now read `TRACE`/`TRACE`, `WARNING`/`DEBUG` and `INFO`/`DEBUG`; the third one used to be `null`.
can iterate if you have more thoughts

PR #584, opened 2026-08-10
Add more MAST physics variables

### LINKED ISSUE #527: Add additional physics methods for MAST
A first pass at adding MAST data was added in #488. However it would be good to add some additional physics methods that are currently missing for consistency:

- [ ] `z_cur` - estimated vertical position of the centroid
- [ ] `z_prog` - programmed vertical position of the centroid
- [ ] `z_error` - error between actual and programmed
- [ ] `n_equal_1_mode` - RMS amplitude of n=1 mode
- [ ] `ne_peaking` - Thomson density peaking factor
- [ ] `te_peaking` - Thomson temperature peaking factor
- [ ] `pressure_peaking` - Thomson pressure peaking factor
- [ ] `te_width` - Thomson width from Gaussian fit
- [ ] `p_rad_peaking` - radiated power peaking factor

samueljackson92:
This adds additional physics variables for:

 - z_cur - estimated vertical position of the centroid
 - z_prog - programmed vertical position of the centroid
- z_error - error between actual and programmed
- ne_peaking - Thomson density peaking factor
- te_peaking - Thomson temperature peaking factor
- pressure_peaking - Thomson pressure peaking factor
- p_rad_peaking - radiated power peaking factor
- te_width - Thomson width from Gaussian fit

This PR closes #527.

### DIFF
--- a/disruption_py/core/utils/misc.py
+++ b/disruption_py/core/utils/misc.py
@@ -17,6 +17,8 @@ from typing import Dict, List, Tuple, Type
 import numpy as np
 from loguru import logger
 
+from disruption_py.core.physics_method.errors import CalculationError
+
 
 def without_duplicates(lst: List):
     """
@@ -207,6 +209,39 @@ def to_tuple(
     Dict[str, Tuple[str, np.ndarray]]
     """
     return {k: (dim, v) for k, v in data.items()}
+
+
+def require_aligned(path: str, signal: np.ndarray, time: np.ndarray) -> np.ndarray:
+    """
+    Check a signal is sampled on the time base it will be interpolated against.
+
+    The last axis is the time axis for both 1-D traces and 2-D (channel, time)
+    profiles, so this covers both.
+
+    Parameters
+    ----------
+    path : str
+        Path of the signal, used for the error message.
+    signal : np.ndarray
+        The signal to check.
+    time : np.ndarray
+        The time base the signal should be sampled on.
+
+    Returns
+    -------
+    np.ndarray
+        The signal, unchanged.
+
+    Raises
+    ------
+    CalculationError
+        If the signal's time axis does not match the length of the time base.
+    """
+    if signal.shape[-1] != len(time):
+        raise CalculationError(
+            f"{path} has {signal.shape[-1]} samples but its time base has {len(time)}"
+        )
+    return signal
 
 
 def filter_dict(i: Dict, s: str) -> Dict:
--- a/disruption_py/core/utils/misc.py
+++ b/disruption_py/core/utils/misc.py
@@ -17,7 +17,7 @@ from typing import Dict, List, Tuple, Type
 import numpy as np
 from loguru import logger
 
-from disruption_py.core.physics_method.errors import CalculationError
+from disruption_py.core.physics_method.errors import MismatchCalculationError
 
 
 def without_duplicates(lst: List):
@@ -211,21 +211,16 @@ def to_tuple(
     return {k: (dim, v) for k, v in data.items()}
 
 
-def require_aligned(path: str, signal: np.ndarray, time: np.ndarray) -> np.ndarray:
+def assert_equal_length(a: np.ndarray, b: np.ndarray) -> np.ndarray:
     """
-    Check a signal is sampled on the time base it will be interpolated against.
-
-    The last axis is the time axis for both 1-D traces and 2-D (channel, time)
-    profiles, so this covers both.
+    Assert that two arrays have the same length, raising an error if they do not.
 
     Parameters
     ----------
-    path : str
-        Path of the signal, used for the error message.
-    signal : np.ndarray
-        The signal to check.
-    time : np.ndarray
-        The time base the signal should be sampled on.
+    a : np.ndarray
+        The first array to check.
+    b : np.ndarray
+        The second array to check.
 
     Returns
     -------
@@ -234,14 +229,14 @@ def require_aligned(path: str, signal: np.ndarray, time: np.ndarray) -> np.ndarr
 
     Raises
     ------
-    CalculationError
-        If the signal's time axis does not match the length of the time base.
+    MismatchCalculationError
+        If the lengths of the two arrays do not match.
     """
-    if signal.shape[-1] != len(time):
-        raise CalculationError(
-            f"{path} has {signal.shape[-1]} samples but its time base has {len(time)}"
+    if a.shape[-1] != len(b):
+        raise MismatchCalculationError(
+            f"a has length {a.shape[-1]} but b has length {len(b)}"
         )
-    return signal
+    return a
 
 
 def filter_dict(i: Dict, s: str) -> Dict:
--- a/disruption_py/inout/xr.py
+++ b/disruption_py/inout/xr.py
@@ -48,7 +48,7 @@ class ProcessXarrayConnection(ProcessConnection):
     @property
     def folder_path(self):
         """Get full folder path including endpoint URL if provided."""
-        if self.endpoint_url is not None:
+        if self.endpoint_url:
             return f"{self.endpoint_url}/{self.file_path}"
 
         return self.file_path
--- a/disruption_py/inout/xr.py
+++ b/disruption_py/inout/xr.py
@@ -137,9 +137,6 @@ class XarrayDataConnection(DataConnection):
         try:
             item = self.data_tree[path]
         except KeyError as e:
-            if not required:
-                nan_result = np.array([np.nan])
-                return xr.DataArray(nan_result) if return_xarray else nan_result
             raise FetchDataError(path) from e
 
         if required and not np.any(np.isfinite(item)):
--- a/disruption_py/inout/xr.py
+++ b/disruption_py/inout/xr.py
@@ -137,6 +137,9 @@ class XarrayDataConnection(DataConnection):
         try:
             item = self.data_tree[path]
         except KeyError as e:
+            if not required:
+                nan_result = np.array([np.nan])
+                return nan_result if not return_xarray else xr.DataArray(nan_result)
             raise FetchDataError(path) from e
 
         if required and not np.any(np.isfinite(item)):
--- a/disruption_py/inout/xr.py
+++ b/disruption_py/inout/xr.py
@@ -139,7 +139,7 @@ class XarrayDataConnection(DataConnection):
         except KeyError as e:
             if not required:
                 nan_result = np.array([np.nan])
-                return nan_result if not return_xarray else xr.DataArray(nan_result)
+                return xr.DataArray(nan_result) if return_xarray else nan_result
             raise FetchDataError(path) from e
 
         if required and not np.any(np.isfinite(item)):
--- a/disruption_py/machine/mast/efit.py
+++ b/disruption_py/machine/mast/efit.py
@@ -29,6 +29,7 @@ class MastEfitMethods:
         "rmagz": "magnetic_axis_z",
         "tribot": "triangularity_lower",
         "tritop": "triangularity_upper",
+        "volume": "volume",
         "v_loop_dynamic": "vloop_dynamic",
         "v_loop_static": "vloop_static",
         "wmhd": "wmhd",
--- a/disruption_py/machine/mast/efit.py
+++ b/disruption_py/machine/mast/efit.py
@@ -4,6 +4,8 @@
 Module for retrieving and processing EFIT parameters for MAST.
 """
 
+import numpy as np
+
 from disruption_py.core.physics_method.decorator import physics_method
 from disruption_py.core.physics_method.params import PhysicsMethodParams
 from disruption_py.machine.mast.util import MastUtilMethods
@@ -49,15 +51,33 @@ class MastEfitMethods:
         Returns
         -------
         dict
-            A dictionary containing the retrieved EFIT parameters.
+            A dictionary containing the retrieved EFIT parameters. Properties whose
+            underlying signal is missing for this shot are returned as NaN rather
+            than dropping the whole equilibrium reconstruction.
         """
-        eq_time = params.get_data("equilibrium/time")
         times = params.times
+        eq_time = MastUtilMethods.get_data_or_none(params, "equilibrium/time")
+        if eq_time is None:
+            params.logger.warning(
+                "get_efit_parameters: equilibrium/time not available. Returning NaNs."
+            )
+            return {
+                key: np.full_like(times, np.nan)
+                for key in MastEfitMethods.efit_properties
+            }
 
         outputs = {}
         for key, prop in MastEfitMethods.efit_properties.items():
-            signal = params.get_data(f"equilibrium/{prop}")
-            item = MastUtilMethods.interpolate_1d(eq_time, signal, times)
-            outputs[key] = item
+            signal = MastUtilMethods.get_data_or_none(params, f"equilibrium/{prop}")
+            if signal is None:
+                params.logger.warning(
+                    "get_efit_parameters: equilibrium/{prop} not available. "
+                    "Returning NaNs for {key}.",
+                    prop=prop,
+                    key=key,
+                )
+                outputs[key] = np.full_like(times, np.nan)
+                continue
+            outputs[key] = MastUtilMethods.interpolate_1d(eq_time, signal, times)
 
         return outputs
--- a/disruption_py/machine/mast/efit.py
+++ b/disruption_py/machine/mast/efit.py
@@ -4,10 +4,7 @@
 Module for retrieving and processing EFIT parameters for MAST.
 """
 
-import numpy as np
-
 from disruption_py.core.physics_method.decorator import physics_method
-from disruption_py.core.physics_method.errors import DataError
 from disruption_py.core.physics_method.params import PhysicsMethodParams
 from disruption_py.machine.mast.util import MastUtilMethods
 from disruption_py.machine.tokamak import Tokamak
@@ -57,10 +54,7 @@ class MastEfitMethods:
             than dropping the whole equilibrium reconstruction.
         """
         times = params.times
-        try:
-            eq_time = params.get_data("equilibrium/time", required=True)
-        except DataError:
-            return {key: [np.nan] for key in MastEfitMethods.efit_properties}
+        eq_time = params.get_data("equilibrium/time", required=True)
 
         outputs = {}
         for key, prop in MastEfitMethods.efit_properties.items():
--- a/disruption_py/machine/mast/efit.py
+++ b/disruption_py/machine/mast/efit.py
@@ -7,6 +7,7 @@ Module for retrieving and processing EFIT parameters for MAST.
 import numpy as np
 
 from disruption_py.core.physics_method.decorator import physics_method
+from disruption_py.core.physics_method.errors import DataError
 from disruption_py.core.physics_method.params import PhysicsMethodParams
 from disruption_py.machine.mast.util import MastUtilMethods
 from disruption_py.machine.tokamak import Tokamak
@@ -56,28 +57,14 @@ class MastEfitMethods:
             than dropping the whole equilibrium reconstruction.
         """
         times = params.times
-        eq_time = MastUtilMethods.get_data_or_none(params, "equilibrium/time")
-        if eq_time is None:
-            params.logger.warning(
-                "get_efit_parameters: equilibrium/time not available. Returning NaNs."
-            )
-            return {
-                key: np.full_like(times, np.nan)
-                for key in MastEfitMethods.efit_properties
-            }
+        try:
+            eq_time = params.get_data("equilibrium/time", required=True)
+        except DataError:
+            return {key: [np.nan] for key in MastEfitMethods.efit_properties}
 
         outputs = {}
         for key, prop in MastEfitMethods.efit_properties.items():
-            signal = MastUtilMethods.get_data_or_none(params, f"equilibrium/{prop}")
-            if signal is None:
-                params.logger.warning(
-                    "get_efit_parameters: equilibrium/{prop} not available. "
-                    "Returning NaNs for {key}.",
-                    prop=prop,
-                    key=key,
-                )
-                outputs[key] = np.full_like(times, np.nan)
-                continue
+            signal = params.get_data(f"equilibrium/{prop}")
             outputs[key] = MastUtilMethods.interpolate_1d(eq_time, signal, times)
 
         return outputs
--- a/disruption_py/machine/mast/physics.py
+++ b/disruption_py/machine/mast/physics.py
@@ -864,7 +864,6 @@ class MastPhysicsMethods:
         disruption-py/blob/matlab/CMOD/matlab-core/get_TS_data_cmod.m), adapted from
         the vertical C-Mod geometry to the radial MAST geometry.
         """
-        nan_result = {"te_width": [np.nan]}
 
         # 2D profile array: dims (major_radius, time)
         te_xr = params.get_data("thomson_scattering/t_e", return_xarray=True)
@@ -872,12 +871,7 @@ class MastPhysicsMethods:
         r_ts = te_xr.coords["major_radius"].values
         ts_time = te_xr.coords["time"].values
 
-        try:
-            _, r_mag, a_minor = MastUtilMethods.thomson_rho(params, r_ts)
-        except CalculationError:
-            # equilibrium unavailable: can't bound the Gaussian fit rejection
-            # window, so the width can't be computed for this shot
-            return nan_result
+        _, r_mag, a_minor = MastUtilMethods.thomson_rho(params, r_ts)
 
         return MastPhysicsMethods._get_te_width(
             params.times,
--- a/disruption_py/machine/mast/physics.py
+++ b/disruption_py/machine/mast/physics.py
@@ -278,15 +278,7 @@ class MastPhysicsMethods:
 
         """
 
-        try:
-            n_e = params.get_data("summary/line_average_n_e", required=True)
-        except DataError:
-            # some shots lack a line-average density measurement entirely
-            return {
-                "n_e": [np.nan],
-                "dn_dt": [np.nan],
-                "greenwald_fraction": [np.nan],
-            }
+        n_e = params.get_data("summary/line_average_n_e", required=True)
         t_n = params.get_data("summary/time")
         ip = params.get_data("summary/ip")
         t_ip = params.get_data("summary/time")
@@ -382,35 +374,24 @@ class MastPhysicsMethods:
             A dictionary containing SXR data (`sxr_data`) and
             corresponding time points (`sxr_time`).
         """
-        try:
-            hcam = params.get_data(
-                "soft_x_rays/horizontal_cam_upper",
-                return_xarray=True,
-                required=True,
-            )
-        except DataError:
-            hcam = None
+        hcam = params.get_data(
+            "soft_x_rays/horizontal_cam_upper",
+            return_xarray=True,
+            required=True,
+        )
 
-        if hcam is not None:
-            hcam_channel = hcam.isel(horizontal_cam_upper_channel=0)
-            hcam_channel = hcam_channel.squeeze(drop=True)
-            sxr_time = hcam_channel.time.values
-            sxr_core = hcam_channel.values
-        else:
-            sxr_time = np.array([np.nan])
-            sxr_core = np.array([np.nan])
+        hcam_channel = hcam.isel(horizontal_cam_upper_channel=0)
+        hcam_channel = hcam_channel.squeeze(drop=True)
+        sxr_time = hcam_channel.time.values
+        sxr_core = hcam_channel.values
 
         times = params.times
         sxr_core = MastUtilMethods.interpolate_1d(sxr_time, sxr_core, times)
 
-        if hcam is not None:
-            hcam_channel = hcam.isel(horizontal_cam_upper_channel=7)
-            hcam_channel = hcam_channel.squeeze(drop=True)
-            sxr_edge = hcam_channel.values
-        else:
-            sxr_edge = np.array([np.nan])
+        hcam_channel = hcam.isel(horizontal_cam_upper_channel=7)
+        hcam_channel = hcam_channel.squeeze(drop=True)
+        sxr_edge = hcam_channel.values
 
-        times = params.times
         sxr_edge = MastUtilMethods.interpolate_1d(sxr_time, sxr_edge, times)
 
         return {"sxr_core": sxr_core, "sxr_edge": sxr_edge}
@@ -435,14 +416,11 @@ class MastPhysicsMethods:
             A dictionary containing D-alpha signal data (`dalpha`).
         """
 
-        try:
-            dalpha = params.get_data(
-                "spectrometer_visible/filter_spectrometer_dalpha_voltage",
-                return_xarray=True,
-                required=True,
-            )
-        except DataError:
-            return {"d_alpha": [np.nan]}
+        dalpha = params.get_data(
+            "spectrometer_visible/filter_spectrometer_dalpha_voltage",
+            return_xarray=True,
+            required=True,
+        )
 
         dalpha = dalpha.isel(dalpha_channel=2)
         dalpha = dalpha.dropna(dim="time")
@@ -495,12 +473,8 @@ class MastPhysicsMethods:
         - Rea et al. (2020), *Fusion Sci. Technol.* 76(8), 912-924.
           DOI: 10.1080/15361055.2020.1798589
         """
-        try:
-            power = params.get_data("bolometer/power", required=True)
-            rmag = params.get_data("equilibrium/magnetic_axis_r", required=True)
-        except DataError:
-            # bolometer diagnostic not installed/recorded for this shot
-            return {"prad_peaking": [np.nan]}
+        power = params.get_data("bolometer/power", required=True)
+        rmag = params.get_data("equilibrium/magnetic_axis_r", required=True)
 
         bolo_time = params.get_data("bolometer/time")
         first_r = params.get_data("bolometer/first_point_r")
@@ -720,15 +694,12 @@ class MastPhysicsMethods:
         }
 
         # 2D profile arrays: dims (major_radius, time)
-        try:
-            te_xr = params.get_data(
-                "thomson_scattering/t_e", return_xarray=True, required=True
-            )
-            ne_xr = params.get_data(
-                "thomson_scattering/n_e", return_xarray=True, required=True
-            )
-        except DataError:
-            return nan_result
+        te_xr = params.get_data(
+            "thomson_scattering/t_e", return_xarray=True, required=True
+        )
+        ne_xr = params.get_data(
+            "thomson_scattering/n_e", return_xarray=True, required=True
+        )
 
         try:
             pe_xr = params.get_data(
@@ -903,20 +874,11 @@ class MastPhysicsMethods:
             A dictionary containing the z parameters: `z_error`,
             `z_prog`, `zcur`, `v_z`, and `z_times_v_z`.
         """
-        nan_result = {
-            col: [np.nan] for col in ("z_error", "z_prog", "zcur", "v_z", "z_times_v_z")
-        }
-
-        try:
-            z_ref = params.get_data("pulse_schedule/z_ref", required=True)
-            zip_prx = params.get_data("controllers/zip_proxy", required=True)
-            t_ctrl = params.get_data("controllers/time", required=True)
-            ip_raw = params.get_data("summary/ip", required=True)
-            t_ip = params.get_data("summary/time", required=True)
-        except DataError:
-            # e.g. older shots without the zip_proxy controller signal, or
-            # signal(s) present but entirely NaN
-            return nan_result
+        z_ref = params.get_data("pulse_schedule/z_ref", required=True)
+        zip_prx = params.get_data("controllers/zip_proxy", required=True)
+        t_ctrl = params.get_data("controllers/time", required=True)
+        ip_raw = params.get_data("summary/ip", required=True)
+        t_ip = params.get_data("summary/time", required=True)
 
         ip_ctrl = MastUtilMethods.interpolate_1d(t_ip, ip_raw, t_ctrl)
 
--- a/disruption_py/machine/mast/physics.py
+++ b/disruption_py/machine/mast/physics.py
@@ -11,6 +11,7 @@ import scipy.constants as const
 from disruption_py.core.physics_method.decorator import physics_method
 from disruption_py.core.physics_method.errors import (
     CalculationError,
+    FetchDataError,
     MismatchCalculationError,
 )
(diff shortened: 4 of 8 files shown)


gtrevisan:
@gtrevisan
- functional

@samueljackson92
- after revision: run `disruption_errors`

@AlexSaperstein @nbarbour13
- ne_peaking - Thomson density peaking factor
- te_peaking - Thomson temperature peaking factor
- pressure_peaking - Thomson pressure peaking factor
- te_width - Thomson width from Gaussian fit

@zapatace @nbarbour13 
- p_rad_peaking - radiated power peaking factor
- z_cur - estimated vertical position of the centroid
- z_prog - programmed vertical position of the centroid
- z_error - error between actual and programmed
- current_quench_time tweaks?

disruption_py/machine/mast/efit.py
```diff
@@ -48,15 +51,33 @@ def get_efit_parameters(params: PhysicsMethodParams):
...
         -------
         dict
-            A dictionary containing the retrieved EFIT parameters.
+            A dictionary containing the retrieved EFIT parameters. Properties whose
+            underlying signal is missing for this shot are returned as NaN rather
+            than dropping the whole equilibrium reconstruction.
         """
-        eq_time = params.get_data("equilibrium/time")
         times = params.times
+        eq_time = MastUtilMethods.get_data_or_none(params, "equilibrium/time")
```
gtrevisan, line 59: I believe this should be enough to assert that `eq_time` is not full of NaNs:
```suggestion
        eq_time = params.get_data("equilibrium/time", required=True)
```

I introduced `required` here:
https://github.com/MIT-PSFC/disruption-py/blob/901684c55ce20a33f7f6c151dde15cf26f8568c9/disruption_py/inout/xr.py#L142-L143

furthermore, our custom errors will already be downplayed as warnings, here:
https://github.com/MIT-PSFC/disruption-py/blob/901684c55ce20a33f7f6c151dde15cf26f8568c9/disruption_py/core/physics_method/runner.py#L189-L190

disruption_py/machine/mast/efit.py
```diff
@@ -48,15 +51,33 @@ def get_efit_parameters(params: PhysicsMethodParams):
...
+                key: np.full_like(times, np.nan)
+                for key in MastEfitMethods.efit_properties
+            }
 
         outputs = {}
         for key, prop in MastEfitMethods.efit_properties.items():
-            signal = params.get_data(f"equilibrium/{prop}")
-            item = MastUtilMethods.interpolate_1d(eq_time, signal, times)
-            outputs[key] = item
+            signal = MastUtilMethods.get_data_or_none(params, f"equilibrium/{prop}")
```
gtrevisan, line 71: here you wouldn't use `required`, but rather return what you have.
you *can* use a single `[np.nan]` instead of a properly-sized array, as the framework will take care of it, here:
https://github.com/MIT-PSFC/disruption-py/blob/901684c55ce20a33f7f6c151dde15cf26f8568c9/disruption_py/core/physics_method/runner.py#L265-L269

what was the error you were trying to downplay to warning, here?

disruption_py/machine/mast/physics.py
```diff
@@ -41,19 +46,29 @@ def get_ip_parameters(params: PhysicsMethodParams):
...
-        dipprog_dt = np.gradient(ip_prog, ip_prog_time)
 
         times = params.times
 
         ip = MastUtilMethods.interpolate_1d(magtime, ip, times)
-        ip_prog = MastUtilMethods.interpolate_1d(ip_prog_time, ip_prog, times)
         dip_dt = MastUtilMethods.interpolate_1d(magtime, dip_dt, times)
-        dipprog_dt = MastUtilMethods.interpolate_1d(ip_prog_time, dipprog_dt, times)
+
+        ip_prog = MastUtilMethods.get_data_or_none(params, "pulse_schedule/i_plasma")
```
gtrevisan, line 58: if I understand what you would like to accomplish, I'm not sure you:
1. really want to suppress any error -- how would the user notice these missing data points?
2. really need a custom MAST method -- why not simply catch what you need?

disruption_py/machine/mast/physics.py
```diff
@@ -78,20 +96,109 @@ def get_power(params: PhysicsMethodParams):
...
         -------
         dict
-            A dictionary containing neutral beam injection power (`power_nbi`) and
-            radiated power (`power_radiated`).
+            A dictionary containing neutral beam injection power (`p_nbi`),
+            total Ohmic heating power (`p_oh`), and radiated power (`p_rad`).
+
+        References
+        ------
+        - pull requests: #[553](https://github.com/MIT-PSFC/disruption-py/pull/553)
```
gtrevisan, line 104: I guess you can now add *this* PR to the references...

disruption_py/machine/mast/physics.py
```diff
@@ -78,20 +96,109 @@ def get_power(params: PhysicsMethodParams):
...
+                "Returning NaNs for p_rad."
+            )
+            power_radiated = np.full_like(times, np.nan)
+
+        try:
+            power_ohm = MastPhysicsMethods._get_p_ohm(params)
+        except (CalculationError, FetchDataError) as exc:
+            # a missing equilibrium or Ip signal must not take out p_nbi and p_rad
+            params.logger.warning("p_oh: {exc}", exc=exc)
+            power_ohm = np.full_like(times, np.nan)
```
gtrevisan, line 138: as a general rule of thumb, yes, sometimes you might want to set your data vars to NaNs, but in that case for simplicity please use the short-hand:
```suggestion
            power_ohm = [np.nan]
```

disruption_py/machine/mast/physics.py
```diff
@@ -237,12 +399,24 @@ def _get_densities(times, n_e, t_n, ip, t_ip, a_minor, t_a):
...
+        MastUtilMethods.require_aligned("ip", ip, t_ip)
         # get the gradient of n_E
         dn_dt = np.gradient(n_e, t_n)
         n_e = interp1(t_n, n_e, times)
         dn_dt = interp1(t_n, dn_dt, times)
         ip = -ip / 1e6  # Convert from A to MA and take positive value
         ip = interp1(t_ip, ip, times)
+
+        if not np.any(np.isfinite(a_minor)) or np.size(t_a) < 2:
+            # No usable equilibrium: the densities stand, the Greenwald fraction cannot.
```
gtrevisan, line 412: you might want to drop a warning yourself, here.

disruption_py/machine/mast/physics.py
```diff
@@ -319,15 +521,323 @@ def get_dalpha(params: PhysicsMethodParams):
         dalpha_data = MastUtilMethods.interpolate_1d(dalpha_time, dalpha_data, times)
         return {"d_alpha": dalpha_data}
 
+    @staticmethod
+    def _get_prad_peaking(
```
gtrevisan, line 525: it *is* true that C-mod at present still shows separation of fetching vs computation routines in public vs private methods, but I'm not sure that is helping out, at all...
leave it if you think this is useful, but otherwise feel free to reabsorb these in a single public method.

disruption_py/machine/mast/physics.py
```diff
@@ -319,15 +521,323 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        efit_time : np.ndarray
+            Time base of the equilibrium data.
+
+        Returns
+        -------
+        dict
+            A dictionary containing `prad_peaking`.
+        """
+        if power.ndim != 2:
+            raise CalculationError(
```
gtrevisan, line 577:
```suggestion
            raise MismatchCalculationError
```

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +847,263 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
+        # 2D profile array: dims (major_radius, time)
+        te_xr = MastUtilMethods.get_data_or_none(
+            params, "thomson_scattering/t_e", return_xarray=True
+        )
+
+        if te_xr is None:
+            params.logger.warning(
+                "get_te_width: thomson_scattering/t_e not available. Returning NaNs."
+            )
+            return nan_result
```
gtrevisan, line 1024: as a general rule of thumb, I'd keep things easy and avoid catching/logging/returning within the physics method if that's what the framework would have done if you had not handled the exception in the first place... assuming you wanted to use the `required=True` arg, of course.

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +847,263 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
+        r_ts = te_xr.coords["major_radius"].values
+        ts_time = te_xr.coords["time"].values
+
+        try:
+            _, r_mag, a_minor = MastUtilMethods.thomson_rho(params, r_ts)
+        except CalculationError as exc:
+            # equilibrium unavailable: can't bound the Gaussian fit rejection
+            # window, so the width can't be computed for this shot
+            params.logger.warning("get_te_width: {exc}", exc=exc)
+            return nan_result
```
gtrevisan, line 1035: again, wouldn't it be the same if you didn't caught the `CalculationError`?

disruption_py/machine/mast/util.py
```diff
@@ -18,42 +22,150 @@ class MastUtilMethods:
     """
 
     @staticmethod
-    def retrieve_ip(conn: XarrayDataConnection):
+    def get_data_or_none(params: PhysicsMethodParams, path: str, **kwargs):
```
gtrevisan, line 25: I'm not sure this would be needed at the backend level.
could you share an example log of what you are trying to fix, here?

I think it's fair to assume missing data will produce lots of FetchDataError lines.

disruption_py/machine/mast/util.py
```diff
@@ -18,42 +22,150 @@ class MastUtilMethods:
...
+        np.ndarray, xr.DataArray, or None
+            The fetched data, or None if the path does not exist for this shot.
+        """
+        try:
+            return params.get_data(path, **kwargs)
+        except FetchDataError:
+            return None
+
+    @staticmethod
+    def require_aligned(path: str, signal: np.ndarray, time: np.ndarray) -> np.ndarray:
```
gtrevisan, line 56: if you feel this is needed, then we could move it someplace else, eg in `utils/math`...

gtrevisan (changes requested):
functionally, I think we can streamline this a bit.
(I have not commented on every single instance of these issues, so please apply the same pattern everywhere.)

your version with a custom raise for `efit_time`:
```
uv run disruption-py 11867 -t mast -m get_efit_parameters

[WARNING] #11867 | get_efit_parameters: equilibrium/time not available. Returning NaNs.
[ ERROR ] #11867 | Error! No parameters have data: 0/17 (0.00%)
```

my suggestion using `required`:
```
uv run disruption-py 11867 -t mast -m get_efit_parameters

[WARNING] #11867 | get_efit_parameters: FetchDataError('equilibrium/time')
[ ERROR ] #11867 | Error! No parameters have data: 0/17 (0.00%)
```
isn't that effectively the same?

if possible, I'd remove any legwork from the physics methods if the framework manages to do the same.

when you want to fallback and return only *some* data vars, then yes, do not use `required` for those variables.

what do you think?
can you provide some log patterns to show what you wanted to suppress?
thanks!

disruption_py/core/utils/misc.py
```diff
@@ -17,6 +17,8 @@
 import numpy as np
 from loguru import logger
 
+from disruption_py.core.physics_method.errors import CalculationError
```
gtrevisan:
```suggestion
from disruption_py.core.physics_method.errors import MismatchCalculationError
```

disruption_py/core/utils/misc.py
```diff
@@ -209,6 +211,39 @@ def to_tuple(
...
+    np.ndarray
+        The signal, unchanged.
+
+    Raises
+    ------
+    CalculationError
+        If the signal's time axis does not match the length of the time base.
+    """
+    if signal.shape[-1] != len(time):
+        raise CalculationError(
```
gtrevisan:
```suggestion
        raise MismatchCalculationError(
```

disruption_py/core/utils/misc.py
```diff
@@ -209,6 +211,39 @@ def to_tuple(
     return {k: (dim, v) for k, v in data.items()}
 
 
+def require_aligned(path: str, signal: np.ndarray, time: np.ndarray) -> np.ndarray:
```
gtrevisan: let me propose this name change since you are not really checking alignment (meaning values), only length.

```suggestion
def assert_equal_length(path: str, signal: np.ndarray, time: np.ndarray) -> np.ndarray:
```

disruption_py/core/utils/misc.py
```diff
@@ -209,6 +211,39 @@ def to_tuple(
...
+        The signal, unchanged.
+
+    Raises
+    ------
+    CalculationError
+        If the signal's time axis does not match the length of the time base.
+    """
+    if signal.shape[-1] != len(time):
+        raise CalculationError(
+            f"{path} has {signal.shape[-1]} samples but its time base has {len(time)}"
```
gtrevisan: maybe we should be more generic and talk about `a` and `b` as opposed to `signal` vs `time`.

disruption_py/inout/xr.py
```diff
@@ -137,6 +137,9 @@ def get_data(
         try:
             item = self.data_tree[path]
         except KeyError as e:
+            if not required:
```
gtrevisan: as we discussed, I'd remove this:
- KeyErrors will be reraised as FetchDataError
- requirement gets tested only if data is present

disruption_py/machine/mast/efit.py
```diff
@@ -48,15 +52,19 @@ def get_efit_parameters(params: PhysicsMethodParams):
...
-            A dictionary containing the retrieved EFIT parameters.
+            A dictionary containing the retrieved EFIT parameters. Properties whose
+            underlying signal is missing for this shot are returned as NaN rather
+            than dropping the whole equilibrium reconstruction.
         """
-        eq_time = params.get_data("equilibrium/time")
         times = params.times
+        try:
+            eq_time = params.get_data("equilibrium/time", required=True)
+        except DataError:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -78,20 +95,89 @@ def get_power(params: PhysicsMethodParams):
...
+
+        power_radiated = params.get_data("summary/power_radiated")
         power_radiated = MastUtilMethods.interpolate_1d(
             base_time, power_radiated, times
         )
-        return {"p_nbi": power_nbi, "p_rad": power_radiated}
+
+        try:
+            power_ohm = MastPhysicsMethods._get_p_ohm(params)
+        except (CalculationError, FetchDataError):
```
gtrevisan: for brevity, maybe:
```suggestion
        except CustomError:
```

disruption_py/machine/mast/physics.py
```diff
@@ -113,28 +199,23 @@ def get_gas(params: PhysicsMethodParams):
...
 
-        times = params.times
-        total_injected = MastUtilMethods.interpolate_1d(
-            base_time, total_injected, times
-        )
-        inboard_total = MastUtilMethods.interpolate_1d(base_time, inboard_total, times)
-        outboard_total = MastUtilMethods.interpolate_1d(
-            base_time, outboard_total, times
-        )
+        def _fetch(path: str) -> np.ndarray:
```
gtrevisan: although useful, I think we should leave getters and interpolators untouched for now, as we do have plans to take control of this at the framework level in the future.

disruption_py/machine/mast/physics.py
```diff
@@ -194,24 +278,36 @@ def get_densities(params: PhysicsMethodParams):
 
         """
 
-        n_e = params.get_data("summary/line_average_n_e")
+        try:
+            n_e = params.get_data("summary/line_average_n_e", required=True)
+        except DataError:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -194,24 +278,36 @@ def get_densities(params: PhysicsMethodParams):
...
         t_a = params.get_data("equilibrium/time")
 
         return MastPhysicsMethods._get_densities(
-            params.times, n_e, t_n, ip, t_ip, a_minor, t_a
+            params, params.times, n_e, t_n, ip, t_ip, a_minor, t_a
         )
 
     @staticmethod
-    def _get_densities(times, n_e, t_n, ip, t_ip, a_minor, t_a):
+    def _get_densities(params, times, n_e, t_n, ip, t_ip, a_minor, t_a):
```
gtrevisan: it might be better to *avoid* following the C-mod example with public/private methods, but you can also leave it as is, your call -- depending on how much time you have. :)

disruption_py/machine/mast/physics.py
```diff
@@ -270,17 +382,38 @@ def get_sxr(params: PhysicsMethodParams):
...
             corresponding time points (`sxr_time`).
         """
-        hcam = params.get_data("soft_x_rays/horizontal_cam_upper", return_xarray=True)
+        try:
+            hcam = params.get_data(
+                "soft_x_rays/horizontal_cam_upper",
+                return_xarray=True,
+                required=True,
+            )
+        except DataError:
```
gtrevisan: from all the checks on whether `hcam` is None or not, this seems to be strictly required for your method. if so:
remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -270,17 +382,38 @@ def get_sxr(params: PhysicsMethodParams):
...
+            hcam = None
+
+        if hcam is not None:
+            hcam_channel = hcam.isel(horizontal_cam_upper_channel=0)
+            hcam_channel = hcam_channel.squeeze(drop=True)
+            sxr_time = hcam_channel.time.values
+            sxr_core = hcam_channel.values
+        else:
```
gtrevisan: `else` looks unnecessary to me.

disruption_py/machine/mast/physics.py
```diff
@@ -270,17 +382,38 @@ def get_sxr(params: PhysicsMethodParams):
...
-        sxr_time = hcam.time.values
-        sxr_data = hcam.values
+        times = params.times
+        sxr_core = MastUtilMethods.interpolate_1d(sxr_time, sxr_core, times)
+
+        if hcam is not None:
+            hcam_channel = hcam.isel(horizontal_cam_upper_channel=7)
+            hcam_channel = hcam_channel.squeeze(drop=True)
+            sxr_edge = hcam_channel.values
+        else:
```
gtrevisan: `else` looks unnecessary to me.

disruption_py/machine/mast/physics.py
```diff
@@ -302,10 +435,14 @@ def get_dalpha(params: PhysicsMethodParams):
...
-            "spectrometer_visible/filter_spectrometer_dalpha_voltage",
-            return_xarray=True,
-        )
+        try:
+            dalpha = params.get_data(
+                "spectrometer_visible/filter_spectrometer_dalpha_voltage",
+                return_xarray=True,
+                required=True,
+            )
+        except DataError:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -321,13 +458,246 @@ def get_dalpha(params: PhysicsMethodParams):
...
+        -------
+        - Rea et al. (2020), *Fusion Sci. Technol.* 76(8), 912-924.
+          DOI: 10.1080/15361055.2020.1798589
         """
-        Calculate the ohmic heating power from the dynamic loop voltage,
-        inductive voltage, and plasma current
+        try:
+            power = params.get_data("bolometer/power", required=True)
+            rmag = params.get_data("equilibrium/magnetic_axis_r", required=True)
+        except DataError:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -321,13 +458,246 @@ def get_dalpha(params: PhysicsMethodParams):
...
+            "first_point_r": first_r,
+            "first_point_z": first_z,
+            "second_point_r": second_r,
+            "second_point_z": second_z,
+            "validity": validity,
+            "channel_type": channel_type,
+        }
+        for name, array in per_channel.items():
+            if array.shape != (n_channels,):
+                raise CalculationError(
```
gtrevisan, line 487: you might want to use your mismatch utility function (if you made it a bit more generic than signal vs time), or raise a Mismatch error.

disruption_py/machine/mast/physics.py
```diff
@@ -321,13 +458,246 @@ def get_dalpha(params: PhysicsMethodParams):
...
+
+        Returns
+        -------
+        dict
+            A dictionary containing `te_peaking`, `ne_peaking` and
+            `pressure_peaking`.
+        """
+        core_mask = (
+            rho < 0.3
+        )  # channels within 30% of normalized radius (Rea et al. 2020, Eq. 7)
```
gtrevisan: at this point I'd add the comment on a single line before a single line of code -- black will format it like this otherwise.

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +707,233 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
+            `pressure_peaking`.
 
         References
-        ------
-        - pull requests: #[553](https://github.com/MIT-PSFC/disruption-py/pull/553)
+        -------
+        - Rea et al. (2020), *Fusion Sci. Technol.* 76(8), 912-924.
+          DOI: 10.1080/15361055.2020.1798589
         """
+        nan_result = {
```
gtrevisan: you should never need this.

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +707,233 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
-        equilibrium_time = params.get_data("equilibrium/time")
+        # 2D profile arrays: dims (major_radius, time)
+        try:
+            te_xr = params.get_data(
+                "thomson_scattering/t_e", return_xarray=True, required=True
+            )
+            ne_xr = params.get_data(
+                "thomson_scattering/n_e", return_xarray=True, required=True
+            )
+        except DataError:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +707,233 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
-            dip_smoothed = dip_dt
-        inductance = 4.0 * np.pi * 1.0e-7 * r0 * li / 2.0
+            pe_profile = pe_xr.values
+
+        r_ts = te_xr.coords["major_radius"].values
+        ts_time = te_xr.coords["time"].values
+
+        try:
+            rho, _, _ = MastUtilMethods.thomson_rho(params, r_ts)
+        except CalculationError:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +707,233 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
+            col: [np.nan] for col in ("z_error", "z_prog", "zcur", "v_z", "z_times_v_z")
+        }
+
+        try:
+            z_ref = params.get_data("pulse_schedule/z_ref", required=True)
+            zip_prx = params.get_data("controllers/zip_proxy", required=True)
+            t_ctrl = params.get_data("controllers/time", required=True)
+            ip_raw = params.get_data("summary/ip", required=True)
+            t_ip = params.get_data("summary/time", required=True)
+        except DataError:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +707,233 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
+        params : PhysicsMethodParams
+            The parameters containing the Xarray connection, shot id and more.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the z parameters: `z_error`,
+            `z_prog`, `zcur`, `v_z`, and `z_times_v_z`.
+        """
+        nan_result = {
```
gtrevisan: you should never need this.

disruption_py/machine/mast/util.py
```diff
@@ -18,42 +22,87 @@ class MastUtilMethods:
...
+
+        Raises
+        ------
+        CalculationError
+            If the equilibrium magnetic axis or minor radius is unavailable.
+        """
+        try:
+            r_mag = params.get_data("equilibrium/magnetic_axis_r", required=True)
+            a_minor = params.get_data("equilibrium/minor_radius", required=True)
+        except DataError as exc:
```
gtrevisan: remove the try/except block -- the framework will do just the same if you don't handle the exceptions.

gtrevisan (commented):
I did my best to go through this again after our quick chat.
LMK if you notice any weird pattern that could be improved from the physics methods -- considering that we'd want to do that for *all* physics methods, though.

after some thought, I'd avoid your get_and_interpolate (`_fetch`) utility function as we should revamp the interpolation in the (medium) future, but okay if you'd prefer to leave it.

disruption_py/machine/mast/efit.py
gtrevisan: you forgot about this one. 🚧

disruption_py/machine/mast/physics.py
```diff
@@ -41,19 +48,26 @@ def get_ip_parameters(params: PhysicsMethodParams):
...
         ip = MastUtilMethods.interpolate_1d(magtime, ip, times)
-        ip_prog = MastUtilMethods.interpolate_1d(ip_prog_time, ip_prog, times)
         dip_dt = MastUtilMethods.interpolate_1d(magtime, dip_dt, times)
-        dipprog_dt = MastUtilMethods.interpolate_1d(ip_prog_time, dipprog_dt, times)
+
+        try:
+            ip_prog = params.get_data("pulse_schedule/i_plasma", required=True)
+            ip_prog_time = params.get_data("pulse_schedule/time", required=True)
+        except DataError:
+            # older shots lack a recorded pulse schedule; ip/dip_dt stand on their own
```
gtrevisan, line 63: since you are now handling the `DataError`, you may want to drop a log line about this so the user knows *why* those data vars are empty.

disruption_py/machine/mast/physics.py
```diff
@@ -270,17 +347,27 @@ def get_sxr(params: PhysicsMethodParams):
...
+            return_xarray=True,
+            required=True,
+        )
 
-        hcam = hcam.isel(horizontal_cam_upper_channel=7)
-        hcam = hcam.squeeze(drop=True)
-        hcam = hcam.drop_vars(["horizontal_cam_upper_channel"])
-        sxr_time = hcam.time.values
-        sxr_data = hcam.values
+        hcam_channel = hcam.isel(horizontal_cam_upper_channel=0)
```
gtrevisan: FYI -- you've updated the channel from 7 to 0, is this on purpose?

disruption_py/machine/mast/physics.py
```diff
@@ -270,17 +347,27 @@ def get_sxr(params: PhysicsMethodParams):
...
+        hcam_channel = hcam_channel.squeeze(drop=True)
+        sxr_time = hcam_channel.time.values
+        sxr_core = hcam_channel.values
 
         times = params.times
-        sxr_data = MastUtilMethods.interpolate_1d(sxr_time, sxr_data, times)
-        return {"sxr": sxr_data}
+        sxr_core = MastUtilMethods.interpolate_1d(sxr_time, sxr_core, times)
+
+        hcam_channel = hcam.isel(horizontal_cam_upper_channel=7)
```
gtrevisan: back to 7 here. again -- FYI.

disruption_py/machine/mast/physics.py
gtrevisan: I guess you might be doing something slightly different here, so feel free to resolve this comment.

disruption_py/machine/mast/physics.py
```diff
@@ -337,42 +582,202 @@ def get_ohmic_parameters(params: PhysicsMethodParams):
...
+        )
 
-        # compute derived quantities
-        smooth_window_size = 30
-        dip_dt = np.gradient(ip, summary_time)
-        if len(dip_dt) >= smooth_window_size:
-            dip_smoothed = causal_boxcar_smooth(dip_dt, smooth_window_size)
+        te_profile = te_xr.values
+        ne_profile = ne_xr.values
+        if pe_xr is None:
```
gtrevisan: now that you are using `required=True`, I don't think `pe_xr` can be `None`.

gtrevisan (commented):
I gave it another go -- this looks much better!

gtrevisan (commented):
thanks, @samueljackson92, LGTM! let's wait for a physics-based review.

nbarbour13:
I took a look at the `te_width` method for the set of test shots you included in the `config.toml`.
The code looks good, and the results below look reasonable.  

Each plot shows the Level 2 Thomson T_e data pulled from the archive.
The overplotted black lines have length `te_width` calculated by your Disruption-Py physics method.
Each line segment is centered at the magnetic axis position, pulled from the MAST Level 2 `equilibrium` data.

mast_te_width_updated

<br>
<br>

The anomalous value around t=0.05 ms for shot 29666 can be partially explained by an additional plot, illustrating that at an early time for one of these shots, the Gaussian temperature assumption is invalid:
image

The call to `scipy.optimize.curve_fit` completes without failure on the t=0.05 time slice, despite the non-Gaussian distribution.
The result is that the `te_width` probably should not be used for that time slice and others like it.
As your code is an implementation of the existing `te_width` method for the other machines, I vote that we include your method as-is.


Alex Saperstein suggested that perhaps we could include a note to users in metadata as a reminder that the Gaussian assumption that the `te_width` relies on is not always satisfied.  I'm open to your thoughts as we review metadata for the upcoming release.

AlexSaperstein:
I reviewed the Thomson peaking method/features, including ne_peaking, te_peaking, and pressure_peaking.

tldr: the method and features look good, with some caveats

1. the calcs may not be as accurate early on in the discharge (~O(50 ms) into the shot). the solution to this would introduce a lot of logistical complexities
2. the local/core-average in rho-space may not be the most appropriate for spherical tokamak. so take absolute values with caution. especially when comparing them to values that come from conventional aspect ratio tokamaks using the same calculation.

More detailed investigation:

Looking at each qualitative aspects of each on a small scale (~2 shots each) shows that the peaking factors are following the change in peaking over time relatively appropriately.

ne_peaking
image
image

- higher ne_peaking clearly corresponds to more peaked profiles

However, some shots reviewed highlight that these calculations may not be accurate during the early stages of the discharge (~O(50 ms) into the shot)

te_peaking & pressure peaking
image
image

- this is not really unexpected. the issue is clearly just that the peak in the ne/te/pe is clearly off-centered at these times, leading to a severe underestimation of the peaking factor.
- this shift in the profiles is very likely the result of using \<Rmagx\> (the time-average of the major radius position of the magnetic axis) to map R -> rho-space for the calculation.
https://github.com/MIT-PSFC/disruption-py/blob/e76983566c67a2e6c6c20e66da4217e71f357b6c/disruption_py/machine/mast/util.py#L97
- This assumption is okay when Rmagx ~ \<Rmagx\>, which is typically the case for most of the shot. but early on in the shot, Rmagx << \<Rmagx\> is common, as the the plasma starts off inner-wall limited.
- Updating the R->rho mapping to use the instantaneous Rmagx rather than \<Rmagx\> would resolve this issue, but I can understand the logistical issue associate with then having a rho(time) coordinate.
- This early inaccuracy is probably fine for now, it should just be kept in mind when using this feature.

Something else I noticed when investing the R->rho mapping is that the local/core averaging of ne/te/pressure in rho-space may be more inappropriate for a spherical tokamak like MAST than it is for a conventional aspect ratio tokamak like C-Mod or DIII-D.
- The reason for this is that spherical tokamaks like MAST tend to have much larger Shafranov shifts (due to their higher betas and lower aspect ratios)
- The notable size of the Shafranov shift is reflected in the O(15%) of the minor radius (i.e. "rho") difference between the geometric and magnetic axes that I've seen for the shots I've inspected.
- 15% is a significant shift in the context of these peaking calculations, as the "core" is defined as the inner 30% of the plasma in rho-space. This can result in the HFS of the average only encompassing as much as ~50% of the "core" as it should (due to the increased spacing between flux surfaces on the HFS caused by the Shafranov shift, and the fact that the most accurate peaking-factor calculation would use psi to define the core, rather than rho, as rho is NOT a flux coordinate)
- I did not go too far down the rabbit hole of quantifying how much this might affect the calculated peaking factors, but it should be kept in mind that the calculation returned here might not have absolute magnitudes consistent with those for conventional aspect ratio tokamaks (or in general tokamaks with smaller Shafranov shifts)

AlexSaperstein (approved):
See previous post

gtrevisan:
I'll let @samueljackson92 read/reply to the points above, if necessary.

status update, AFAIK:

@gtrevisan
- [x] functional

@samueljackson92
- [ ] after revision: run `disruption_errors`

@AlexSaperstein @nbarbour13
- [x] ne_peaking - Thomson density peaking factor
- [x] te_peaking - Thomson temperature peaking factor
- [x] pressure_peaking - Thomson pressure peaking factor
- [x] te_width - Thomson width from Gaussian fit

@zapatace @nbarbour13 
- [ ] p_rad_peaking - radiated power peaking factor
- [ ] z_cur - estimated vertical position of the centroid
- [ ] z_prog - programmed vertical position of the centroid
- [ ] z_error - error between actual and programmed
- [ ] current_quench_time tweaks?

zapatace:
I downloaded 500 random MAST shots and checked: current_quench_time (revisited), p_rad_peaking, z_cur, z_prog 
and z_error.

In the errors log these methods raise only `FetchDataError` (missing signal). `get_current_quench_time` raises nothing.

### `current_quench_time`

Most current quench times are correct, I found only 5 shots that genuinely missed their current quench time computation (there are more, but they look like rampdowns or calibration shots). It is important to note that most shots end with a small QC, even at the end of what it looks like ramp-downs so this field should not be used to detect disruptivity in MAST. 

For example, you have here some ramp-down behaviour ( Red dashed line = `current_quench_time`):
image

 
and one of the missed shots...
image

Some disruptions are rejected because current gets too negative (−57 kA in this case). `abs_ip_final` is greater than |50 kA| so it is not flagged. We might need to change this abs for machines with this negative current behaviour.

Since this method is generic, I would like to build a multimachine dataset before we modify/tweak some core component... Specific rules can be added to MAST though.

### `prad_peaking`

402/500 shots have data (92 lack `bolometer/power`). Most look physical: 0.5–0.8 in ramp-up, 1–1.7 in flat-top.

- 40/402 shots exceed 10, and 9 have > 5 % negative values. In these shots 43–50 % of chord samples are negative, so the bin means become small or negative. `abd_mean != 0` does not catch this.
- C-Mod's `get_prad_peaking` sets negative brightness to NaN before averaging. You can see next example of applying this guard.

image

The flat part after 0.21 s follows an unphysical 50 MW `p_rad` spike (stuck bolometer signal), which the guard does not address. 

### `zcur`, `z_prog`, `z_error`

The main problem is that only 253/500 shots have data (see stats below), apparently because controllers/zip_proxy  exists from shot ~21475. Any explanation for this?

- **`z_prog`** (`pulse_schedule/z_ref`) exists in 498/500 shots but is dropped with the rest because `zip_proxy` is `required`. Could it be returned on its own?
- **Physics looks right.** In flat-top |z_error| has a median of 0.2 cm (90 % < 1.3 cm), and it grows in the ms before the quench (VDE, below).
- **Noise for t < 0** Before breakdown |Ip| exceeds 10 kA in 170/500 shots, so `zcur`, `z_error` and `v_z` are noise for t < 0 and after the quench. Should we increase the threshold beyond 10kA?  

Here you have an example (`z_error` rises before the quench. The noise at t < 0 passes the 10 kA guard.).
image



### `Stats`

Here you can find availability of the different columns:

image

As mentioned before, we have a significant fraction of missing data for some fields. 

And next we have distributions, which look healthy, maybe we can run a follow up study about outliers beyond 1 and 99 percentiles. 

image

samueljackson92:
Thank you all for the hard work in validating & checking the new physics methods for MAST. I very much appreciate it!

@nbarbour13 and @AlexSaperstein my take away from your review is that the implementation is sensible and consistent with C-Mod, with the noted caveats. I guess it would be worth discussing at the next meeting whether there is anything we can/should sensibly try to do to mitigate those caveats?

@zapatace:

1. Good that the CQ times appear to be mostly correct. 👍 Maybe we can try tweak thresholds a bit more, but I am not sure how much more we'd gain...
2. P_rad peaking: great comments. I think we should obviously make the suggested changes for negative values. Maybe we can also check for stuck bolometer signals..
3. Yes good point about the `zip_proxy`. Looking at the UDA logs it looks like it was not defined before 21305, which is likely causing a lot of the missing values.

image

nbarbour13:
That sounds like a good topic for discussion.  I agree with Alex that mitigating some of the concerns about early times may require a cumbersome implementation. My vote is to discuss avenues for improvement in our next meeting and to still merge the Thomson-based methods as you have implemented them.

Thanks for the great reviews, Enrique and Alex, and thank you for the work you did to write these methods, Samuel!

gtrevisan:
as we discussed in our last meeting, the target here is a great bang-for-buck review, leaving further tweaks to be tackled separately.

@samueljackson92 @nbarbour13 @AlexSaperstein @zapatace
where do we stand on this?

Sam pushed one commit to take care of nonphysical negative values.
are we waiting for more tweaks, or further review?
or may I merge? thank you all!

nbarbour13 (approved):
From my perspective we're good to merge, and we can fix remaining issues as users report them.

disruption_py/machine/mast/physics.py
```diff
@@ -78,20 +98,85 @@ def get_power(params: PhysicsMethodParams):
...
         power_radiated = MastUtilMethods.interpolate_1d(
             base_time, power_radiated, times
         )
-        return {"p_nbi": power_nbi, "p_rad": power_radiated}
+
+        try:
+            power_ohm = MastPhysicsMethods._get_p_ohm(params)
+        except CustomError:
+            # a missing equilibrium or Ip signal must not take out p_nbi and p_rad
+            power_ohm = [np.nan]
```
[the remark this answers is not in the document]
samueljackson92: @gtrevisan I believe we agreed that in dpy framework we should use `[np.nan]` rather than `full_like` and let the framework handle it.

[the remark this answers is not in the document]
gtrevisan: yes, agreed!

PR #587, opened 2026-08-18
Unload datasets by default

gtrevisan:
we want `DictOutputSetting`, which is our under-the-hood *de facto* base class, to be memory-efficient and therefore more reliable for large-scale workflows.

once a `Pool`-processed dataset is fed to `output_shot`:
- save it (`Dict`) or shard it (subclasses) to disk,
- unload it from memory,
- lazy reload it,

furthermore, for the (abstract) subclass `SingleOutputSetting`:
- when getting the results:
  - explicit read before concatenation,
- when saving to disk:
  - removed superseded shards.

as a consequence:
- all workflows collect shards, even in fatal cases (eg OOM killed, SLURM time limits, any framework-level unhandled exception)
  - `DictOutputSetting` in the target folder,
  - `SingleOutputSetting` in the temporary folder,
- `DictOutputSetting` should complete the workflow with a low memory footprint, namely just enough to have lazy-loaded datatasets as dictionary values,
- by design, as before, `DictOutputSetting` requires explicit downstream concatenation.
- `SingleOutputSetting` should work effectively as before, although the new logic is to unload each process results to disk and to load them up again from disk just before concatenation. this results in a minor performance hit which is negligible due to the no-encoding and no-compression defaults of xarray's `to_netcdf`, and which is overshadowed by the newly-introduced impossibility of losing data during the concatenation stage,
- new debug statements are now tracking the maximum RSS memory (with a high-watermark level!) throughout the workflow, including after each shot dataset is populated,
- new renaming logic is now avoiding clashes about pre-existing same-name files, and still saving data to disk.

the option of invoking an `OutputSetting` with `path=False` should work as before, but:
- `DictOutputSetting`: it defeats the purpose of memory optimization as it cannot lazy reload any dataset,
- `SingleOutputSetting`: it will still save, reload, load, concat, and delete shards, but any exception before a successful `to_disk` call will leave data behind -- by design! to always be able to recover individual shards. eg: `DatasetOutputSetting` workflow that takes 12h and dies due to OOM right as it's trying to concatenate all results into its single dataset.

example runs for 1,000 C-MOD shots:

- `DatasetOutputSetting`:
```
Starting workflow: MaxRSS = 201.0 MB
Completed workflow: MaxRSS = 729.0 MB
Read shots: MaxRSS = 1,231.0 MB
Concatenated shots: MaxRSS = 2,555.4 MB
Saved to disk: MaxRSS = 2,555.4 MB
```

- `DictOutputSetting`:
```
Starting workflow: MaxRSS = 197.0 MB
Completed workflow: MaxRSS = 719.0 MB
Saved to disk: MaxRSS = 731.0 MB
# custom post-processed load step to compare final footprint
Loaded: MaxRSS = 1,227.0 MB
```

I already have a plan to develop a new `OutputSetting` to efficiently concat large-scale workflow results.
@samc24 thoughts on the architecture? for sure we could change many things, including fully revamping the multiprocessing data exchange and all the output settings, but this felt a great bang-for-the-buck improvement to me.
@nbarbour13 tests should already pass, but feel free to play around with the various output settings to make sure the results still end up where they are supposed to, and no shards get left behind.


disruption_py/settings/output_setting.py
```diff
@@ -178,7 +195,43 @@ def _output_shot(self, params: OutputSettingParams):
...
+        else:
+            # shard into temporary folder
+            fd, shard = tempfile.mkstemp(
+                dir=get_temporary_folder(), prefix=f".{params.shot_id}.", suffix=".nc"
+            )
+            os.close(fd)
+
+        # save to disk
+        self.shards[params.shot_id] = shard
+        logger.trace(
```
nbarbour13: I think increasing the visibility of this from `trace` to `debug` is appropriate.
My preference would be for users to be informed at the `debug` level at minimum whenever files are written to disk.

Since the shard files are removed on successful execution, I agree that it is fair to lower the visibility from `info`.

disruption_py/settings/output_setting.py
```diff
@@ -266,45 +315,65 @@ def get_results(self) -> OutputSingleType:
...
+                raise NotImplementedError("Could not save object to file.")
+            logger.info(
+                "Saved {type} in {took:.3f} s: {path}",
+                type=type(self).__name__.removesuffix("OutputSetting"),
+                took=time.time() - t,
+                path=self.path,
+            )
+
+        for shard in self.shards.values():
+            logger.trace("Removing shard: {shard}", shard=shard)
```
nbarbour13: Same as above: I think this should be raised to `debug` visibility.

nbarbour13 (changes requested):
I tested this branch on the C-Mod test shots using the default settings and log level set to `trace`.  Shards were removed as expected.  Manually raising an exception at line 374 in the `to_disk` method of the `SingleOutputSetting` properly blocks the cleanup.  The data in the remaining shard were complete.  

I request that a change be made to raise the visibility of actions that write or delete temporary data to disk.  

Overall this is a good addition to the codebase that will enable robust generation of large datasets.

gtrevisan:
as discussed in the meetings, at this time I'd keep the logic as is, with a clear preference for data recoverability over disk usage and unnecessary unloading/loading of single-output subclasses.

if needed, we can allow fully-disk-less workflows (as before) in a separate PR.
maybe we'll get some further CS improvements by @samc24 in the near future.

otherwise, this PR should be ready to go. :crossed_fingers:

nbarbour13 (approved):
This looks good to me.  I agree with the idea to revisit a disk-less option in the future.
Low priority: documentation should be updated to reflect the new write paradigm.

samc24 (changes requested):
- Fair reasoning to keep the logic as is, will take a look the CS follow-up lol 
- Went through the current head with a few synthetic runs of the output settings (no tokamak data), and one thing is maybe worth fixing before merge; the rest I would scope into the follow-up PR.

**Before merge.** `OutputSettingList(["dataset", "datatree"])` constructs both members when the default `output.nc` does not exist yet, so both get the same path and the second `to_disk` overwrites the first. Verified in a fresh process on this branch: the surviving file opens as the DataTree, the Dataset is gone. But the same case on dev branch raises `FileExistsError` at the second write, so this is a regression from loud to silent. I would think that reserving the default path with `mkstemp` at construction, which is what the old pytest branch effectively did, would fix it?

**Follow-up PR, what I would scope in:**
- Shard once per shot above the output settings (base `output_shot` or the workflow loop) instead of per member. Removes the double write for two single-output members in a list, and sets up writing the shard in the worker and returning the path instead of the dataset, which kills the pickle round trip through the Pool pipe and parallelizes the netcdf write.
- Replace the `type(self) is DictOutputSetting` dispatch with a hook the subclass overrides (where shards go, whether `to_disk` removes them). Today a trivial `class MyDict(DictOutputSetting): pass` takes the subclass branch: hidden shards in the temp folder, no output folder, a "Saved results" log for a folder that does not exist, and no cleanup. Nobody does this today, but subclassing is the documented extension point.
- Return the output paths alongside the results. With rename-on-collision the caller asked for `results.nc` and the data is in `results.abc123.nc`, and since `get_shots_data` drops the return value of `to_disk`, the log is the only way to learn the real path.
- A shard-lifecycle test in `test_output_setting.py` (which files exist after `output_shot`, `get_results`, `to_disk` per setting type and for `path=False`). Needs no tokamak data.
- Docs for the new write paradigm, per @nbarbour13's note :D

gtrevisan:
> **Before merge.** `OutputSettingList(["dataset", "datatree"])` constructs both members when the default `output.nc` does not exist yet, so both get the same path and the second `to_disk` overwrites the first. Verified in a fresh process on this branch: the surviving file opens as the DataTree, the Dataset is gone. But the same case on dev branch raises `FileExistsError` at the second write, so this is a regression from loud to silent. I would think that reserving the default path with `mkstemp` at construction, which is what the old pytest branch effectively did, would fix it?

I've just pushed a minimum-effort fix:
- as soon as you establish that eg `output.nc` does not exist, `touch` it,
- explicit `INFO`-level log about destination folder or file.

I believe it should resolve the race condition both you and Copilot flagged.
in any case, `OutputSettingList` exists only for testing purposes, and I'd be fine dropping it.
no user will ever use this combo OutputSetting -- at best, they will extract one OutputSetting and then manipulate it as a downstream postprocessing step.

> * Shard once per shot above the output settings (base `output_shot` or the workflow loop) instead of per member. Removes the double write for two single-output members in a list, and sets up writing the shard in the worker and returning the path instead of the dataset, which kills the pickle round trip through the Pool pipe and parallelizes the netcdf write.

you write "pickle" to mean memory, here, or is there a behind-the-scenes pickle serialization happening on disk?
in any case, yes, that's what I had in mind at first when I embarked in this feature PR, but since the `output_shot` method is there precisely to "stream" data to disk serially from the main process, I thought it best to keep the pattern as-is and simplify my development.
we will discuss in person the priority of this improvement, but I've already collected thoughts from our group showing that I'm basically the only one running 60,000-shot workflows, and most people love exchanging memory as opposed to loading files.
(also, let's keep in mind that a typical DIII-D worker node has 512 GB of memory but only 320 GB of local scratch disk.)

> * Replace the `type(self) is DictOutputSetting` dispatch with a hook the subclass overrides (where shards go, whether `to_disk` removes them). Today a trivial `class MyDict(DictOutputSetting): pass` takes the subclass branch: hidden shards in the temp folder, no output folder, a "Saved results" log for a folder that does not exist, and no cleanup. Nobody does this today, but subclassing is the documented extension point.

let's pin this for the future.
I have a separate sibling repo (currently private, but public in the near future) where I developed what will very likely become a `ZarrOutputSetting` which subclasses `DictOutputSetting` and then manually and efficiently concatenates everything together without ever loading all data in memory at the same time:
- https://github.com/MIT-PSFC/disruption-db

> * Return the output paths alongside the results. With rename-on-collision the caller asked for `results.nc` and the data is in `results.abc123.nc`, and since `get_shots_data` drops the return value of `to_disk`, the log is the only way to learn the real path.

I thought about that.
for now, if it's any consolation, the actual files are included in the xarray Datasets in `ds.encoding["source"]`.
while that would be a good addition, I also want to think through the case in which `DictOutputSetting` finds a filename collision in an already-existing folder and choose to rename some shard -- in such a way, you have the final folder containing two shards for possibly the same shot, which might lead to confusion or even exceptions upon concatenation, eg through `glob.glob`.

> * A shard-lifecycle test in `test_output_setting.py` (which files exist after `output_shot`, `get_results`, `to_disk` per setting type and for `path=False`). Needs no tokamak data.

sure, but the test would need to inspect its own files, and possibly clean up after itself, and so on.
based on my previous experience with `pytest`, its `conftest` module, and all the module vs function scoping of the fixtures, etc, I'd consider this overkill, all the more if in the future we choose to have only Dict sharding to disk.

> * Docs for the new write paradigm, per @nbarbour13's note :D

we could volunteer Nathaniel for that! :smile:
but let's keep working towards our full data release, as APS-DPP is approaching... fast. :rabbit2: 
for now, I hope the docstrings and the new log levels will suffice.

samc24 (approved):
cool, tested new push in fresh processes and the List overwrite is gone. both members get distinct files, and after `to_disk` one opens as the Dataset and the other as the DataTree.

one side effect: the touch happens right before the `os.path.exists` check, so every default-path run now takes the collision branch. since `"dataset"` is the default for both the CLI and `get_shots_data`, a plain `disruption-py -l info <shot>` prints `WARNING Output file already exists! .../output.nc`, saves to `output.<random>.nc`, and leaves an empty `output.nc` next to it. same for the csv default. at least from what i saw . CI does not see it because every test passes an explicit path. the mkstemp name also comes with 0600 permissions, which may matter for outputs on shared disks.

this kept the clean name for me (single default stays `output.nc` with no warning, second List member gets renamed):
```
    default = path is True
    if default:
        ext = "csv" if isinstance(self, DataFrameOutputSetting) else "nc"
        path = os.path.join(get_temporary_folder(), f"output.{ext}")

    if default and not os.path.exists(path):
        # reserve the default path
        with open(path, "x", encoding="utf8"):
            pass
    elif path and os.path.exists(path):
        ...
```
not blocking from my side, approving. if you do drop `OutputSettingList`, the list option is still described in docs/usage/settings/output_setting.md.

on pickle: in memory only, nothing on disk. the worker returns the Dataset through `pool.imap`, so it gets pickled through the pipe and unpickled in the main process, which then writes the shard. point taken on who actually runs 60k-shot workflows hahaha

on paths: `encoding["source"]` works for the Dict output, but for a Dataset result it points at the first temporary shard, which `to_disk` has already removed, so the log is still the only place to find the real file there. good point on renamed shards ending up as two files for the same shot in one folder, I had not considered that. fine to leave it for now.

fine to pin the hook until ZarrOutputSetting lands since that would be the first real subclass, and ok on skipping the test and docs for now.

gtrevisan:
> one side effect: the touch happens right before the `os.path.exists` check, so every default-path run now takes the collision branch. since `"dataset"` is the default for both the CLI and `get_shots_data`, a plain `disruption-py -l info <shot>` prints `WARNING Output file already exists! .../output.nc`, saves to `output.<random>.nc`, and leaves an empty `output.nc` next to it. same for the csv default. at least from what i saw . CI does not see it because every test passes an explicit path. the mkstemp name also comes with 0600 permissions, which may matter for outputs on shared disks.

my bad! hopefully my further commits fix this. (and folder creation.)
but I still haven't tested it -- that's how much I dislike the `OutputSettingList`. :smile: 

> this kept the clean name for me (single default stays `output.nc` with no warning, second List member gets renamed):

I saw you added code after I pushed mine, sorry.

> on pickle: in memory only, nothing on disk. the worker returns the Dataset through `pool.imap`, so it gets pickled through the pipe and unpickled in the main process, which then writes the shard.

alright, so here "pickle" means "serialization", got it.

> on paths: `encoding["source"]` works for the Dict output, but for a Dataset result it points at the first temporary shard, which `to_disk` has already removed, so the log is still the only place to find the real file there.

wow, I did not know that -- it is indeed unexpected as those Datasets are created through `xr.concat` so they are technically new objects which should not carry over any of those shard files' memory/existence.
might it be a bug? I certainly cannot go in the `encoding` structure and tweak the `source` myself. :face_with_diagonal_mouth:

samc24:
cool cool i tested new changes the same way as before: small hand-built datasets pushed through `output_shot` / `get_results` / `to_disk`, each case in a fresh process, listing the files after each step, plus one real CLI run (MAST, 1 shot, default output). all good now

for `encoding["source"]` that seems to be how xarray does concat... not a bug i think. the result takes its encoding from the first dataset , so it reports whichever shard comes first. if I reverse the concat order it reports shard 2 instead. and you can tweak it, `encoding` is a plain dict, so `self.result.encoding["source"] = self.path` in `to_disk` works and the data stays readable. for the DataTree the root has no source and each child carries its own shard path. your call whether it is worth a line, happy to leave it. 🚀 

> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

hahaha they cut you off

gtrevisan:
> you can tweak it, `encoding` is a plain dict, so `self.result.encoding["source"] = self.path` in `to_disk` works and the data stays readable

what do you mean with "stays" readable? it sounds like it's unreadable if you do not tweak the source.
the concat yields a fully-loaded dataset so the data should be all in memory -- nothing needs to be lazy-loaded from disk, so that parameter should be moot.
I will leave this xarray "bug" as is, _unless_ you experienced any data-from-nonexistent-file shenanigan! let me know.

> hahaha they cut you off

that's weird because I am currently at ~75% my monthly credit limit, maybe there is a separate cutoff for reviews only? mysteries of the octocat.

the only other issue I was unsure about was the usage of `resource` to pull some memory utilization stats without further packages (eg `psutil`). I've seen OS reporting wildly different usages than `resources`, so now I'm evaluating a better route. thoughts?

gtrevisan:
I've implemented a lower-level utility to read both the current (`rss`) and peak (`hwm`) RSS usage as the current `psutil` provides only the former, and the `psutil` vs `resources` comparison was off.
we might want to simply pull `psutil` starting with v8+.
EDIT: it's linux only for now!

samc24:
bad wording on "stays readable", no shenanigans 😅 . checked latest without touching `encoding`: after `to_disk` removes the shards the result is plain numpy in memory for all three types, values right, copy / arithmetic / re-save all fine. the stale `source` is just a label, agreed to leave it.

for memory i ran a quick comparison on mfews04 (allocate 800 MB, free it, then print all three in the main process and in fork/spawn workers). in the same process `ru_maxrss` == `VmHWM` and psutil's rss == `VmRSS` to within rounding, so if your numbers looked wildly different i suspect it was peak vs current? (after freeing, peak says 814, current says 14). one thing did look off though: in a spawn worker `ru_maxrss` came back as 814 MB while `VmHWM` said 16, so the worker was reporting the parent's peak. fork workers were fine. took me a bit of digging to understand why, but from what i can tell it's because python spawns via vfork+exec on linux, and the child inherits the parent's high-water mark through that. i checked by launching the child with plain fork+exec instead and the number goes back to normal. either way `VmHWM` is per-process so it doesn't have the problem, which makes /proc the safer route for per-worker numbers, and probably matters more once spawn/forkserver is the default.

two nits: on macOS `get_rss()` returns (0.0, 0.0), so the debug lines say `RSS = 0.0 MB, MaxRSS = 0.0 MB` like a real measurement (the macos CI job too). `resource` can still give the peak there (bytes, like the old code handled), current isn't available without psutil, so maybe fall back for the peak or skip the line. and `-> float` returns a tuple now. psutil would be a new runtime dep btw, it's only in the lock through the optional lab group.

gtrevisan:
> for memory i ran a quick comparison on mfews04 (allocate 800 MB, free it, then print all three in the main process and in fork/spawn workers). in the same process `ru_maxrss` == `VmHWM` and psutil's rss == `VmRSS` to within rounding, so if your numbers looked wildly different i suspect it was peak vs current? (after freeing, peak says 814, current says 14). one thing did look off though: in a spawn worker `ru_maxrss` came back as 814 MB while `VmHWM` said 16, so the worker was reporting the parent's peak. fork workers were fine. took me a bit of digging to understand why, but from what i can tell it's because python spawns via vfork+exec on linux, and the child inherits the parent's high-water mark through that. i checked by launching the child with plain fork+exec instead and the number goes back to normal. either way `VmHWM` is per-process so it doesn't have the problem, which makes /proc the safer route for per-worker numbers, and probably matters more once spawn/forkserver is the default.

I've also seen differences between the HWM reported in `/proc/self/status` (eg by future `psutil` releases post PR 2737)) and that reported from `resources`, so I've chosen to opt for reading both from the former.
(furthermore, I haven't investigated that much on the issue of 'peak' vs 'HWM'.)
we can discuss this further in person and tweak it in another PR, if necessary.

this comment, taken once again from `psutil`'s PR 2737, might provide some insight, too:

> ```
> # VmHWM (from /proc/pid/status) and ru_maxrss both
> # track peak RSS but are synced independently. Allow 5%
> # tolerance.
> ```

> two nits: on macOS `get_rss()` returns (0.0, 0.0), so the debug lines say `RSS = 0.0 MB, MaxRSS = 0.0 MB` like a real measurement (the macos CI job too). `resource` can still give the peak there (bytes, like the old code handled), current isn't available without psutil, so maybe fall back for the peak or skip the line. and `-> float` returns a tuple now. psutil would be a new runtime dep btw, it's only in the lock through the optional lab group.

macOS support for now is only to be taken at face value through the CI jobs, as I don't have a mac system to develop and prototype workflows.
the easiest fireproof way for me to silence that error was setting both out as zeros.

samc24:
LGTM, ship it  🚢

PR #592, opened 2026-09-02
Set variables with mismatched lengths to nan and include in data dict for output.

nbarbour13:
# Problem
While preparing the Disruption-DB for release, Gregorio noticed that DIII-D Shot 203598 was missing `ip_prog_rt` in the output NetCDF file. In debugging, I identified that the try-except block that reads this signal successfully reads the raw signal but fails when subsequently reading the Ip polarity from PTDATA.  Critically, this occurs before interpolation.  The result is that `ip_prog_rt` returns, but it is not interpolated onto the requested time base for the shot.
Further investigation revealed the omission of data variables with mismatched time bases.  An error is logged, but the variable is not written to the output file.

# Solution
Modified the `populate_shot` method in `disruption_py/core/physics_method/runner.py` to add variables with mismatched lengths to the `dict` that becomes `data_vars`.
The method now stores these variables as `nan` with lengths matched to the `physics_method_params` time base.


gtrevisan:
thanks for investigating!
I think we should take the opportunity to look at this a bit better.

is there any salvageable data in the mismatched array, at all?
if so, maybe we should save what we can and return the original data with a bunch of intertwined NaNs.

this portion of code is invoked when the physics method returns a dictionary of numpy arrays, rather than a direct xarray dataset, and so some manipulations are in order -- eg this will be an array of numbers without the corresponding timebase.
by using this bug, or by catching from the logs other similar bugs, we might find some best pattern to be deployed to the physics methods that might make the solution a lot easier at the framework level.

nbarbour13:
Yes, there are many possible ways to resolve this issue, and it is possible that we can find a better solution.
In this instance (DIII-D shot 203598), signal `ipsiptargt` can be read from PTDATA, but the TDI function to read `iptdirect` from PTDATA fails.   

The `get_rt_ip_parameters` physics method uses the 0th element of `iptdirect` to determine the polarity of the current target, and it uses the polarity to process `ip_prog_rt`.
The method could be restructured to include fallback logic for cases where the polarity cannot be accessed from PTDATA.

Some options:
- Implement a fallback to Interpolate the raw Ip target onto the `PhysicsMethodParams` time base, ignoring polarity.
- Create new logic to handle polarity when it is inaccessible and use it to process the Ip target data.
- Return a `nan` array as though the Ip target signal cannot be retrieved. (This is the approach taken by the PR.)

gtrevisan (changes requested):
alright, following our discussion, this is indeed the functional way to go in the framework, coupled with a surgical fix in the offending physics method.

I'd suggest to adjust it technically so that you have something like:
```python
if all(np.isnan(v)):
   logger.debug("all-nan")
else:
   logger.error("length mismatch")
data[k] = np.nan * times
```

gtrevisan (approved):

PR #596, opened 2026-09-03
Fix get_rt_ip_parameters to return ip_prog_rt as nan when PTDATA polarity read fails.

nbarbour13:
# Problem

Related to the discussion in PR #592 , the `get_rt_ip_parameters` physics method returns `ip_prog_rt` in its native time base if reading polarity from PTDATA fails.  Physics methods should return data that has been interpolated onto the `PhysicsMethodParams` time base, or they should return `nan`.  

# Solution

Cleaned up the `except` block to set `ip_prog_rt` to `nan` in accordance with the warning message that is already in the method.

# Comments

The `ip_prog_rt` signal may be salvageable in these cases.  However, this would require developing new logic for handling the polarity in cases where it cannot be read.  While these cases are rare, users of this signal are welcome to suggest solutions for us to revisit this method at a later time.

### DIFF
--- a/disruption_py/machine/d3d/physics.py
+++ b/disruption_py/machine/d3d/physics.py
@@ -809,8 +809,8 @@ class D3DPhysicsMethods:
                 "ip_prog_rt: Failed to get programmed plasma current parameters. Setting to NaN."
             )
             params.logger.opt(exception=True).debug(e)
-            # Return nan if polarity read fails after ip_prog_rt read succeeds.
             ip_prog_rt = [np.nan]
+            dipprog_dt_rt = [np.nan]
         try:
             ip_error_rt, t_ip_error_rt = params.get_data_with_dims(
                 f"ptdata('ipeecoil', {params.shot_id})"
--- a/disruption_py/machine/d3d/physics.py
+++ b/disruption_py/machine/d3d/physics.py
@@ -767,7 +767,8 @@ class D3DPhysicsMethods:
         - original source: [get_Ip_parameters_RT.m](https://github.com/MIT-PSFC/disruption-py
         /blob/matlab/DIII-D/get_Ip_parameters_RT.m)
         - pull requests: #[254](https://github.com/MIT-PSFC/disruption-
-        py/pull/254), #[547](https://github.com/MIT-PSFC/disruption-py/pull/547)
+        py/pull/254), #[547](https://github.com/MIT-PSFC/disruption-py/pull/547),
+        #[596](https://github.com/MIT-PSFC/disruption-py/pull/596)
         - issues: #[506](https://github.com/MIT-PSFC/disruption-py/issues/506)
         """
         ip_rt = [np.nan]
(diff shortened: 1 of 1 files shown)


disruption_py/machine/d3d/physics.py
```diff
@@ -808,6 +809,8 @@ def get_rt_ip_parameters(params: PhysicsMethodParams):
                 "ip_prog_rt: Failed to get programmed plasma current parameters. Setting to NaN."
             )
             params.logger.opt(exception=True).debug(e)
+            # Return nan if polarity read fails after ip_prog_rt read succeeds.
```
gtrevisan: we don't know _which_ line triggered an MdsException, so your comment is technically imprecise, let's just drop it.
the following line is clear and needs no comment.

disruption_py/machine/d3d/physics.py
```diff
@@ -808,6 +809,8 @@ def get_rt_ip_parameters(params: PhysicsMethodParams):
...
+            ip_prog_rt = [np.nan]
```
gtrevisan, line 812: you should probably also set `dipprog_dt_rt` to NaN.

gtrevisan (changes requested):

disruption_py/machine/d3d/physics.py
nbarbour13: All variables returned by the method are initialized to `[np.nan]`.  If an MdsException is raised at line 748, `ip_prog_rt` already returns `[np.nan]` without this PR.  
The only other line where an `MdsException` can occur is line 802, and which is the case where resetting `ip_prog_rt` to `nan` is necessary.

However, as you mention, the except block executes for any MdsException, so I will remove the comment.

disruption_py/machine/d3d/physics.py
nbarbour13: All returned variables are initialized to `[np.nan]`.  `dipprog_dt` always returns `[np.nan]` if an `MdsException` is raised, as the MDSplus calls occur before `dipprog_dt` is modified.

Is your preference to explicitly set all returned variables to `[np.nan]` in each `except` block instead of relying on the initialization? If so, I can edit the other blocks in this method to do so, including this one.

disruption_py/machine/d3d/physics.py
gtrevisan: if `ip_prog_rt` fails you end up in the same `except` block and reset its value to NaN, which it already was, but still, your comment would lead me to believe that that code gets executed _only if_ `polarity` fails, right? better to drop it, as you did. :+1:

disruption_py/machine/d3d/physics.py
gtrevisan: clearly this function, written many moons ago, needs to be revamped, but yeah, I'd say it would be better to set them both to NaN, even if redundant. the next time we go over this we'll probably rewrite it from scratch following better practices.

gtrevisan (approved):

PR #597, opened 2026-09-08
Add a mapping to execute SQL queries

gtrevisan:
forgot about this old branch that might come in handy. eg:

```
# C-MOD
disruption-py 'select shot from summary where shot between 1150805012 and 1150805022;' -l trace

...

[ TRACE ] Executing query: select shot from summary where shot between 1150805012 and 1150805022;
[ INFO  ] Starting workflow: 11 shots / 1 process
```

```
# DIII-D
disruption-py 'select shot from summaries where pulse_length > 10;' -l trace

...

[ TRACE ] Executing query: select shot from summaries where pulse_length > 10;
[ INFO  ] Starting workflow: 199 shots / 1 process
```

### DIFF
--- a/disruption_py/settings/shotlist_setting.py
+++ b/disruption_py/settings/shotlist_setting.py
@@ -13,6 +13,7 @@ from typing import Dict, List, Type, Union
 
 import numpy as np
 import pandas as pd
+from loguru import logger
 
 import disruption_py.data
 from disruption_py.core.utils.enums import map_string_to_enum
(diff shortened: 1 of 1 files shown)


nbarbour13:
I agree that providing CLI access to the `DatabaseShotlistSetting` would be a useful addition to the codebase.  
I also believe that the Copilot recommendation to handle trailing whitespace characters in queries is worth evaluating.

gtrevisan:
did you try this out? if it works like *you* would expect, I'd merge.

I believe whitespace control to be somewhat tricky according to the current setup of cli entrypoints, so I'd leave it as is -- it's supposed to be a tool for power users anyway.

nbarbour13 (approved):
Understood.  Good to merge then!  The tests I tried work as expected.

PR #599, opened 2026-09-09
Tweak imports in generic module to simplify its own import

gtrevisan:
this should allow for an easy import of the generic methods from any machine module:
- lower the machine-specific imports in the generic module to their respective function-level scopes, and 
- disable the respectives pylint codes.


gtrevisan:
earlier I inadvertently merged an initial version of this PR into `dev` after merging:
- #564

so now this PR only provides a small scope refinement of the same logic.

